PR #183, opened 2014-08-27
CDK-619: Implemented read/write support for Pig

joey:
* The patch still needs to move the pig dependencies under
       dependency management in the parent pom.
     * This needs more tests, in particular on error handling.


joey:
Closing until this becomes a priority.

PR #223, opened 2014-09-23
CDK-581: Updates for java8

rdblue:
This PR alters the travis config to test java8 without pushing to master. This will not be part of the commit when merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: java
 jdk:
-  - oraclejdk7
+  - oraclejdk8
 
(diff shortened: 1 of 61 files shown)


rdblue:
I'm going to close this because it appears that we don't need it at the moment.

tomwhite:
We should still try to get this committed sometime, shouldn't we? It needs rebasing and retesting, but the changes here were good overall.

joey:
+1 Java 7 goes EOL soon.

PR #274, opened 2014-12-13
CDK-844: Update delete command to delete views.

rdblue:
The delete command will now detect view URIs and delete just the content
using View#deleteAll (subject to deleteAll's limitations). This is
guarded by a check that validates the view matches the requested URI
because this is destructive. If there is an option in the requested URI
that isn't in the view's URI, then the command fails. This catches typos
in field names that are not recognized as real constraints and not added
to the view.


kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/DeleteCommand.java
```diff
@@ -0,0 +1,107 @@
...
+   * @param view a View's URI
+   * @param requested the requested View URI
+   * @return true if the view's URI and the requested URI match exactly
+   */
+  @VisibleForTesting
+  static boolean viewMatches(URI view, String requested) {
+    // test that the requested options are a subset of the final options
+    Map<String, String> requestedOptions = optionsForUri(URI.create(requested));
+    Map<String, String> finalOptions = optionsForUri(view);
+    for (Map.Entry<String, String> entry : requestedOptions.entrySet()) {
```
joey: Is it possible for finalOptions to contain more options than requestedOptions? If so, is that a problem? I'm pretty sure the answer is no, but I just wanted to go through the thought experiment.

rdblue: Yes, it happens and isn't a problem. The final options will probably contain things like hdfs:host or hdfs:port, which are set automatically. These should be matched only if the original has them.

joey:
+1, ship it!

PR #275, opened 2014-12-17
CDK-677: Add translation from Hive DDL to Schema.

rdblue:
Tables created in Hive using STORED AS AVRO do not have a schema URI or
literal in the table properties. Instead, the schema is generated from
the table DDL when writing. This translates from table DDL if no schema
is set. This also fixes tables created with STORED AS PARQUET.

### DIFF
--- a/kite-data/kite-data-hive/src/main/java/org/kitesdk/data/spi/hive/HiveSchemaConverter.java
+++ b/kite-data/kite-data-hive/src/main/java/org/kitesdk/data/spi/hive/HiveSchemaConverter.java
@@ -1,0 +1,361 @@
+/*
+ * Copyright 2013 Cloudera Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kitesdk.data.spi.hive;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Lists;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.LinkedList;
+import java.util.List;
+import javax.annotation.Nullable;
+import org.apache.avro.Schema;
+import org.apache.hadoop.hive.metastore.api.FieldSchema;
+import org.apache.hadoop.hive.serde2.typeinfo.ListTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.MapTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.StructTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.TypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoFactory;
+import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoUtils;
+import org.apache.hadoop.hive.serde2.typeinfo.UnionTypeInfo;
+import org.codehaus.jackson.node.NullNode;
+import org.kitesdk.compat.DynConstructors;
+import org.kitesdk.compat.DynMethods;
+import org.kitesdk.data.PartitionStrategy;
+import org.kitesdk.data.spi.FieldPartitioner;
+import org.kitesdk.data.spi.SchemaUtil;
+
+public class HiveSchemaConverter {
+
+  private static final DynMethods.StaticMethod primitiveTypeForName =
+      new DynMethods.Builder("getPrimitiveTypeInfo")
+          .impl(TypeInfoFactory.class, String.class)
+          .buildStatic();
+
+  private static final DynMethods.StaticMethod parseTypeInfo =
+      new DynMethods.Builder("getTypeInfoFromTypeString")
+          .impl(TypeInfoUtils.class, String.class)
+          .buildStatic();
+
+  private static Class<?> findTypeInfoClass(String className) {
+    try {
+      return new DynConstructors.Builder(TypeInfo.class)
+          .impl(className).buildChecked().getConstructedClass();
+    } catch (NoSuchMethodException e) {
+      return null;
+    }
+  }
+
+  // TypeInfo classes that may not be present at runtime
+  @VisibleForTesting
+  static final Class<?> charClass = findTypeInfoClass(
+      "org.apache.hadoop.hive.serde2.typeinfo.CharTypeInfo");
+  @VisibleForTesting
+  static final Class<?> varcharClass = findTypeInfoClass(
+      "org.apache.hadoop.hive.serde2.typeinfo.VarcharTypeInfo");
+  @VisibleForTesting
+  static final Class<?> decimalClass = findTypeInfoClass(
+      "org.apache.hadoop.hive.serde2.typeinfo.DecimalTypeInfo");
+
+  private static final ImmutableMap<String, Schema.Type> TYPEINFO_TO_TYPE =
+      ImmutableMap.<String, Schema.Type>builder()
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -148,7 +148,7 @@
     <vers.hbase2>0.98.3-hadoop2</vers.hbase2>
     <vers.hbase-cdh4>0.94.6-cdh${cdh4.version}</vers.hbase-cdh4>
     <vers.hbase-cdh5>0.96.1.1-cdh${cdh5.version}</vers.hbase-cdh5>
-    <vers.hive>0.10.0</vers.hive>
+    <vers.hive>0.13.0</vers.hive>
```
joey: I'm wary of updating the default Hive version so many versions. What changed that requires that we move to 0.13?

kite-data/kite-data-hive/src/main/java/org/kitesdk/data/spi/hive/HiveSchemaConverter.java
```diff
@@ -0,0 +1,357 @@
...
+import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoUtils;
+import org.apache.hadoop.hive.serde2.typeinfo.UnionTypeInfo;
+import org.codehaus.jackson.node.NullNode;
+import org.kitesdk.compat.DynConstructors;
+import org.kitesdk.compat.DynMethods;
+import org.kitesdk.data.PartitionStrategy;
+import org.kitesdk.data.spi.FieldPartitioner;
+import org.kitesdk.data.spi.SchemaUtil;
+
+public class HiveSchemaConverter {
```
joey: This might be a dumb question, but is there not a way to call into a Hive class and have it do this conversion for us? That would guarantee that we're consistent with Hive native Avro tables.

kite-data/kite-data-hive/src/main/java/org/kitesdk/data/spi/hive/HiveSchemaConverter.java
```diff
@@ -0,0 +1,357 @@
...
+    }
+  }
+
+  @VisibleForTesting
+  static Schema optional(Schema schema) {
+    return Schema.createUnion(Lists.newArrayList(NULL, schema));
+  }
+
+  private static String doc(TypeInfo type) {
+    return null;//return "Converted from '" + String.valueOf(type) + "'";
```
joey: Why does this return null?

kite-data/kite-data-hive/src/main/java/org/kitesdk/data/spi/hive/HiveSchemaConverter.java
```diff
@@ -0,0 +1,357 @@
...
+        isOptional ? NULL_DEFAULT : null);
+  }
+
+  @VisibleForTesting
+  static Schema convert(LinkedList<String> path, String name,
+                        TypeInfo type, Collection<String[]> required) {
+    switch (type.getCategory()) {
+      case PRIMITIVE:
+        if (type.getClass() == charClass || type.getClass() == varcharClass) {
+          // this is required because type name includes length
```
joey: Can we capture the field lengths in a schema doc comment somewhere?

rdblue: I don't think we want to. We decided not to add a CHAR or VARCHAR logical type to Avro and Parquet because there's no way of enforcing it and Hive has to assume that the data it reads has no idea that it is a more restrictive SQL type. I'd rather let the SQL engines handle their special types when there isn't any value to support outside.

kite-data/kite-data-hive/src/main/java/org/kitesdk/data/spi/hive/HiveSchemaConverter.java
rdblue: Testing. I'll fix it.

kite-data/kite-data-hive/src/main/java/org/kitesdk/data/spi/hive/HiveSchemaConverter.java
rdblue: There is an implementation in Hive, [TypeInfoToSchema](https://github.com/apache/hive/blob/trunk/serde/src/java/org/apache/hadoop/hive/serde2/avro/TypeInfoToSchema.java), that does it. I'm reluctant to rely on that one because although public, I'm not confident it is considered part of the stable API. Also, we want to have a slightly different schema when writing, which is why can add required fields based on the partition strategy for writes.

pom.xml
rdblue: 0.13.0 is the first version with the CHAR and VARCHAR types. I thought support for those was significant, plus we need to keep up with the upstream libraries better. We already test with 0.12.0-cdh5, so this isn't a huge jump in terms of compatibility.

kite-data/kite-data-hive/src/main/java/org/kitesdk/data/spi/hive/HiveSchemaConverter.java
joey: That makes sense.

kite-data/kite-data-hive/src/main/java/org/kitesdk/data/spi/hive/HiveSchemaConverter.java
joey: That's why I thought it might be useful in a doc, not as something enforced. That way if I'm looking at the schema in Kite I'm not surprised when I reading data in Hive shows truncated values. Then again, you could make the argument hat I should be looking at the Hive schema since it's the read schema being used regardless of what was written.

pom.xml
joey: My concern is with changing the minimum version that we support. I think we still have some users on Hive 0.10/CDH4 and it'd be nice if we could maintain backwards compatibility for them.

kite-data/kite-data-hive/src/main/java/org/kitesdk/data/spi/hive/HiveSchemaConverter.java
rdblue: Oh, as a doc. Yeah, the doc method that is commented out would do that. :)

kite-data/kite-data-hive/src/main/java/org/kitesdk/data/spi/hive/HiveSchemaConverter.java
tomwhite: Why not ask the Hive developers and see if they consider it a part of the public API? Could they provide a public API for this?

My concern is that keeping the code in Kite means it doesn't get updated when Hive changes. (I think Joey was making the same point.)

pom.xml
tomwhite: "STORED AS PARQUET" was added in Hive 0.13, so it makes sense to update the dependency for that reason. "STORED AS AVRO" is in Hive 0.14.

Until we hit 1.0 we should track dependencies more aggressively, IMO. For backwards compatibility in CDH4 - that should be done in CDH itself.

joey: CDH4 didn't bundle Kite, so we'd have to push users to use old versions. CDH5 does bundle Kite, so it makes sense to push compatibility concerns to the CDH maintainers.

kite-data/kite-data-hive/src/main/java/org/kitesdk/data/spi/hive/HiveSchemaConverter.java
rdblue: I don't think that the two would diverge very much, but I doubt we would have the ability to transform into a schema for writing (which requires partition source fields) and for reading.

There's also a concern about supporting older versions that I didn't realize until reading the discussion on upstream Hive versions. If STORED AS AVRO wasn't released until 0.14.0, then we need at least that version because the DDL to schema conversion was added for it.

kite-data/kite-data-hive/src/main/java/org/kitesdk/data/spi/hive/HiveSchemaConverter.java
rdblue: Fixed!

rdblue:
I fixed the docs issue that Joey caught, but I think there is justification for using a DDL to Avro Schema converter in Kite to work with multiple Hive versions (without STORED AS AVRO) and to meet Kite's need for different read and write Schemas.

I've also not bumped the version to 0.14.0. We can do that separately, but this runs as-is and doesn't depend on Hive 0.14.0, so it seems unnecessary. If we want to track upstream versions more closely (and I agree we do) then we should open an issue for it and update more than just Hive.

I think this is ready to go, so please have another look. Thanks!

tomwhite:
This looks good to me.

PR #277, opened 2014-12-22
CDK-854: Use "stdin" as a file source in CLI.

rdblue:


tomwhite:
+1 in the absence of a better solution

PR #282, opened 2014-12-31
CDK-188: Add inputformat format

rdblue:
This is based on CDK-861 until it is merged to avoid unrelated test failures.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/Formats.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/Formats.java
@@ -61,10 +61,18 @@ public class Formats {
       new CompressionType[] { Uncompressed });
 
   /**
-   * Return a {@link Format} for the format name specified. If
-   * {@code formatName} is not a valid name, an IllegalArgumentException is
-   * thrown. Currently the formats <q>avro</q>, <q>csv</q>, and <q>parquet</q>
-   * are supported. Format names are case sensitive.
+   * INPUTFORMAT: a mapreduce InputFormat (read-only).
+   *
+   * @since 0.18.0
+   */
+  public static final Format INPUTFORMAT = new Format("inputformat", Uncompressed,
+      new CompressionType[] { Uncompressed });
+
+  /**
+   * Return a {@link Format} for the format name specified. If {@code formatName}
+   * is not a valid name, an IllegalArgumentException is thrown. Currently the
+   * formats <q>avro</q>, <q>csv</q>, and <q>parquet</q> are supported. Format names are
+   * case sensitive.
    *
    * @since 0.9.0
    * @return an appropriate instance of Format
@@ -79,6 +87,8 @@ public class Formats {
       return JSON;
     } else if (formatName.equals("csv")) {
       return CSV;
+    } else if (formatName.equals("inputformat")) {
+      return INPUTFORMAT;
     } else {
       throw new IllegalArgumentException("Unknown format type: " + formatName);
     }
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/InputFormatReader.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/InputFormatReader.java
@@ -1,0 +1,201 @@
+/*
+ * Copyright 2013 Cloudera Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kitesdk.data.spi.filesystem;
+
+import com.google.common.base.Preconditions;
+import java.io.IOException;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.mapreduce.InputSplit;
+import org.apache.hadoop.mapreduce.Job;
+import org.apache.hadoop.mapreduce.RecordReader;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.apache.hadoop.mapreduce.TaskAttemptID;
+import org.apache.hadoop.mapreduce.lib.input.FileInputFormat;
+import org.kitesdk.compat.Hadoop;
+import org.kitesdk.data.DatasetDescriptor;
+import org.kitesdk.data.DatasetIOException;
+import org.kitesdk.data.DatasetReaderException;
+import org.kitesdk.data.spi.AbstractDatasetReader;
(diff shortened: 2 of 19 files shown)


kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/InputFormatReader.java
```diff
@@ -0,0 +1,201 @@
...
+  public boolean hasNext() {
+    Preconditions.checkState(ReaderWriterState.OPEN.equals(state),
+        "Attempt to read from a file in state:%s", state);
+
+    // the Iterator contract requires that calls to hasNext() not change the
+    // iterator state. calling next() should advance the iterator. however,
+    // this wraps a RecordReader that reuses objects, so advancing in next
+    // after retrieving the key/value pair mutates the pair. this hack is a way
+    // to advance once per call to next(), but do it as late as possible.
+    if (shouldAdvance) {
```
joey: I think this block needs to be synchronized. Otherwise two threads calling hasNext could advance the state twice.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/Main.java
```diff
@@ -99,6 +100,7 @@
     jc.addCommand("csv-schema", new CSVSchemaCommand(console));
     jc.addCommand("csv-import", new CSVImportCommand(console));
     jc.addCommand("obj-schema", new ObjectSchemaCommand(console));
+    jc.addCommand("informat-import", new InputFormatImportCommand(console));
```
joey: I don't like the command name, but I haven't thought of a better one yet.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Formats.java
```diff
@@ -53,10 +53,18 @@ private Formats() {
       new CompressionType[] { Uncompressed });
 
   /**
-   * Return a {@link Format} for the format name specified. If
-   * {@code formatName} is not a valid name, an IllegalArgumentException is
-   * thrown. Currently the formats <q>avro</q>, <q>csv</q>, and <q>parquet</q>
-   * are supported. Format names are case sensitive.
+   * INPUTFORMAT: a mapreduce InputFormat (read-only).
```
joey: What happens if I try to write a dataset that uses the INPUTFORMAT format? Will it just throw an exception telling me it's an invalid format when I try to create the writer?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/InputFormatUtil.java
```diff
@@ -0,0 +1,63 @@
...
+package org.kitesdk.data.spi.filesystem;
+
+import org.apache.hadoop.mapreduce.RecordReader;
+import org.apache.hadoop.mapreduce.lib.input.FileInputFormat;
+import org.kitesdk.compat.DynConstructors;
+import org.kitesdk.data.DatasetDescriptor;
+import org.kitesdk.data.DatasetException;
+
+public class InputFormatUtil {
+  public static final String INPUT_FORMAT_CLASS_PROP = "kite.inputformat.class";
```
joey: Would there be any value in having the format also support a SerDes class? Essentially this format is half of what the read path in Hive offers. That is, you can provide an InputFormat which determines how files are split, but we don't allow anything that changes how records are deserialized other than a transform function during import.

If we did add a SerDes interface, I'd want the target to build Avro records rather than Hive records.

Either way, I don't think we need this for this use case, but it might be worth adding at some point in the future.

joey:
Overall it looks good, I had a few questions listed above.

My only major issue is that this further entangles the core module with MapReduce classes which would make it that much harder to create a core that doesn't include all of those dependencies.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/InputFormatReader.java
rdblue: From the [docs](http://kitesdk.org/docs/current/apidocs/org/kitesdk/data/DatasetReader.html): "Implementations of DatasetReader are not required to be thread-safe"

I'll add the appropriate NotThreadSafe annotation though.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/Main.java
rdblue: Same here. I'm more than happy to change it. Maybe "format-import"?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Formats.java
rdblue: Yes. The writer implementation checks for supported formats and bails if it can't create one:

https://github.com/kite-sdk/kite/blob/master/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemWriter.java#L220

Good catch though, there should be a check in initialize and also one in the partitioned writer.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/InputFormatUtil.java
rdblue: Yes, we will need to do this in the future as well, but I consider it as separate issue. For that, we may want to have a kite-data-hcatalog module.

rdblue:
About the mapreduce classes: Core is already fairly entangled and this would be moved along with the other FS code to a module outside of the API that can depend on MR. There isn't a great way to get around that and provide real access to data locality information (we could plan our own splits, but letting the delegate formats do it is better).

I think the larger issue for separating the API from MR is to fix the MR implementation to avoid delegating to an InputFormat, more like Crunch does.

joey:
That's fair. If we can get to a clean API module and eventually a free core/common module that'd be good.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/Main.java
tomwhite: inputformat-import?

tomwhite:
We discussed this feature in https://issues.cloudera.org/browse/CDK-188, and there was some concern about treating MR input and output formats as Kite formats, since they aren't necessarily readable by other tools (like Hive). Maybe that objection is addressed by only supporting the read path, as you do here.

joey:
I agree it's awkward as a "format", but I view it as part of a more general import feature to read raw data. This is similar to the CSV support we added which also isn't readable by Hive.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/Main.java
joey: What about `raw-import` or `generic-import`? I think the fact that we're using InputFormat as the API for read "raw" data doesn't mean it has to be in the name of the command.

rdblue:
I think this is a must-do for data integration. There needs to be some way of reading things like Sequence files and using Kite on that data. This in particular isn't ideal because we get some unknown object from the InputFormat, but as you noted, I think it is okay as a read-only format.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Formats.java
rdblue: I've added validation checks in partitioned writer and file writer to ensure that a bad format fails in initialize.

rdblue:
I think this is ready to merge after fixing the issue that Joey found. Thanks for the reviews!

tomwhite:
Can we at least change "informat-import"? I would vote for "inputformat-import".

rdblue:
@tomwhite, I've updated the command name to inputformat-import. Anything else we should change before this goes in?

tomwhite:
+1 from me.

PR #283, opened 2015-01-01
CDK-863: Deprecate DatasetWriter flush/sync, add Flushable.

rdblue:
This deprecates the DatasetWriter#flush and #sync methods, which are
replaced by Flushable#flush. Sync needed to call flush to guarantee the
records were durable, but flush was still exposed, requiring callers to
flush and sync, with sync calling flush immediately afterwards. Instead,
the new Flushable interface carries a stronger guarantee that records
have not only been flushed to the underlying buffer, but that they are
persisted to disk also.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/Flushable.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/Flushable.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2013 Cloudera Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kitesdk.data;
+
+/**
+ * A writer that can guarantee data is present on data nodes.
+ * <p>
+ * {@link #flush} has a stronger guarantee than {@link java.io.Flushable#flush}.
+ * When it returns, data already written is flushed to the OS buffers on all
+ * data nodes responsible for replicas.
+ * <p>
+ * Once data has been flushed, it will be tolerant to single-node and rack
+ * failures.
+ */
+@edu.umd.cs.findbugs.annotations.SuppressWarnings(
+    value="NM_SAME_SIMPLE_NAME_AS_INTERFACE",
+    justification="Intended to be a stricter version of the parent interface")
+public interface Flushable extends java.io.Flushable {
+  /**
+   * Ensure that data has been flushed to OS buffers on all replica data nodes.
+   * <p>
+   * This method has a stronger guarantee than {@link java.io.Flushable#flush}.
+   * When this method returns, data already written has been flushed to the OS
+   * buffers on all data nodes responsible for replicas.
+   */
+  @Override
+  void flush();
+}
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetWriter.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetWriter.java
@@ -16,7 +16,6 @@
 package org.kitesdk.data;
 
 import java.io.Closeable;
-import java.io.Flushable;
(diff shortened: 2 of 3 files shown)


kite-data/kite-data-core/src/main/java/org/kitesdk/data/Flushable.java
```diff
@@ -0,0 +1,39 @@
...
+public interface Flushable extends java.io.Flushable {
+  /**
+   * Ensure that data has been written to disk.
+   * <p>
+   * This method has a stronger guarantee than {@link java.io.Flushable#flush}.
+   * When this method returns, data written is persisted to disk, not just an
+   * underlying buffer.
+   */
+  @Override
+  void flush();
```
joey: Do we really need to consolidate `flush()` and `sync()` semantics into a single method?

In the context of Hadoop, `flush()` (technically `hflush()`) guarantees that data has been flushed from the client and sent to buffers on `${dfs.replication}` number of nodes. The `sync()` (really `hsync()`) method goes further and ensures that the data has been written to stable storage on the remote nodes. Some use cases only need the weaker `hflush()` style semantics and don't want to pay the performance penalty of a fully `hsync()`.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/AvroAppender.java
```diff
@@ -80,7 +80,6 @@ public void flush() throws IOException {
   @Override
   public void sync() throws IOException {
     flush();
-    Hadoop.FSDataOutputStream.hsync.invoke(out);
```
joey: Why are we removing the call to `hsync()`?

`hflush()` doesn't guarantee that data has been written to stable storage.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/CSVAppender.java
```diff
@@ -81,7 +81,6 @@ public void flush() throws IOException {
   @Override
   public void sync() throws IOException {
     flush();
-    Hadoop.FSDataOutputStream.hsync.invoke(outgoing);
```
joey: Why are we removing the call to `hsync()`?

`hflush()` doesn't guarantee that data has been written to stable storage.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/AvroAppender.java
rdblue: This was an accident, I did not intend to remove the hsync call.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Flushable.java
rdblue: The pattern in all of our writers is to call flush before calling sync because hsync doesn't necessarily hflush; sync is effectively hflush+hsync. At the very least, we need to be clear that a call to sync is a call to flush so we don't have callers flushing twice. From there, I started to question whether it was worth having two methods for this in the API.

A single method that provides durability is easier to use in terms of both 1) not having to read docs about why there are two and 2) not making a mistake and calling flush when you mean to sync. And, If sync always calls hflush, then how valuable is flush on its own? Execution returns earlier at the cost of not ensuring data is actually on disk, so anyone writing, say, a Flume Sink will won't be able to take advantage of flush because they must call hsync before committing the current transaction.

I think what little utility is gained by exposing hflush alone is outweighed by the confusion of two methods, one of which seems reliable but is really not. Is that reasonable?

tomwhite: > the confusion of two methods, one of which seems reliable but is really not

Reliable is a relative term. Flush (hflush in Hadoop) is reliable in the face of losing power to a rack, but not to a whole data center. Sync (hsync in Hadoop) is reliable in the face of power failure of a whole DC, but not to a DC fire (say). Also, sync is more expensive than flush, and an application may choose flush over sync to trade some reliability for increased throughput. Etc.

So I think we need to expose both flush and sync at the API level in Kite. This can be done through two interfaces, java.io.Flushable and org.kitesdk.data.Syncable:

```java
public interface Syncable {
  void sync() throws IOException;
}
```

Also, we should try out the replacement interfaces to check that they work, pass tests etc before committing this.

joey: +1 to what @tomwhite wrote. Not every application needs sync.

tomwhite:
The interfaces look right to me now. +1

I created https://issues.cloudera.org/browse/CDK-892 as a follow up, and successfully ran all tests with the deprecated methods removed.

PR #284, opened 2015-01-01
CDK-864: Add DatasetRecordException.

rdblue:
This is similar to Avro's AppendWriteException and signals that a write
operation failed because an invalid (or otherwise unusable) record was
passed in, but the writer can still be used.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetRecordException.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetRecordException.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2013 Cloudera Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetRecordException.java
```diff
@@ -0,0 +1,36 @@
...
+package org.kitesdk.data;
+
+/**
+ * A DatasetException that signals that an individual record has failed.
+ * <p>
+ * When writing, this exception indicates that a record could not be written,
+ * but the writer is still valid and ready to write other records.
+ * <p>
+ * When reading, this exception indicates that a record could not be built from
+ * record data, but that reader is still valid and able to read other records.
```
joey: Does this imply the reader would skip the bad bytes and move on to the next record? Do we have a way of doing that yet?

joey:
One question above, but otherwise this is ready to go in.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetRecordException.java
rdblue: This implies, if thrown, that the reader has skipped the bad record. This is for formats like CSV and JSON, where the data may not have a valid record. CSV, for example, might be missing a required field. In that case, we can still read the next record because it might have valid data.

JSON is a similar case, where we could be missing a field or even have the wrong root type, but the iterator provides JsonNode instances, so we can assume that the next one will work.

joey: Makes sense.

joey:
Ship it!

PR #285, opened 2015-01-06
Changed log level and added the word output

tmalaska:


joey:
+1

I just committed this in 0851d03

Thanks @tmalaska!

PR #286, opened 2015-01-08
CDK-871: Expose CSVFileReader in the SPI.

rdblue:
This makes the CSVFileReader public and adds an InputFormat constructor,
rather than just a FileSystem and Path constructor. Because the
InputFormat may not be a FSDataInputStream, this also adds a check when
attempting to use the reader as a Hadoop RecordReader.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/CSVFileReader.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/CSVFileReader.java
@@ -17,6 +17,7 @@
 package org.kitesdk.data.spi.filesystem;
 
 import au.com.bytecode.opencsv.CSVReader;
+import java.io.InputStream;
 import javax.annotation.Nullable;
 import org.apache.hadoop.mapreduce.InputSplit;
 import org.apache.hadoop.mapreduce.RecordReader;
@@ -41,7 +42,7 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.util.NoSuchElementException;
 
-class CSVFileReader<E> extends AbstractDatasetReader<E> {
+public class CSVFileReader<E> extends AbstractDatasetReader<E> {
 
   private static final Logger LOG = LoggerFactory
       .getLogger(CSVFileReader.class);
@@ -57,7 +58,7 @@ class CSVFileReader<E> extends AbstractDatasetReader<E> {
 
   // progress reporting
   private long size = 0;
-  private FSDataInputStream incoming = null;
+  private InputStream incoming = null;
 
   // state
   private ReaderWriterState state = ReaderWriterState.NEW;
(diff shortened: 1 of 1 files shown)


joey:
The patch looks fine, but I think this is another example of where we need a properly defined SPI that includes these kind of classes that aren't meant for general use, but are meant for external consumers.

rdblue:
> The patch looks fine, but I think this is another example of where we need a properly defined SPI that includes these kind of classes that aren't meant for general use, but are meant for external consumers.

I agree.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/CSVFileReader.java
```diff
@@ -87,11 +101,17 @@ public void initialize() {
...
-    } catch (IOException ex) {
-      throw new DatasetReaderException("Cannot open path: " + path, ex);
+    if (incoming != null) {
+      this.size = 1; // avoid divide by 0 errors
+    } else {
+      Preconditions.checkNotNull(fs, "FileSystem cannot be null");
+      Preconditions.checkNotNull(path, "Path cannot be null");
+      try {
+        this.incoming = fs.open(path);
+        this.size = fs.getFileStatus(path).getLen();
```
tomwhite: What if the file length is zero? Wouldn't you still get a divide-by-zero error later?

BTW this is a bug fix that is independent of the issue being fixed here so should be fixed separately - in a different commit, if not a separate JIRA.

tomwhite:
The SPI and the internal implementation classes have been conflated and both are in the spi package. We could move everything that isn't being used externally to an impl package (we already have a precedent for those), then over time add things to spi only when it has been designed with external use in mind. Thoughts?

joey:
I think that's definitely the direction we want to head towards. There's still a conflation of SPI versus limited audience APIs. In my mind, the SPI has the interfaces that are needed to extend Kite with a new format or to make use of a new storage API. Limited audience APIs are APIs that aren't meant for end-users, but are meant for other frameworks that need to integrate Kite, like Flume or Sqoop.

Should we explore an annotation based approach like Hadoop has adopted?

tomwhite:
I have reservations about the Hadoop annotations, since they are widely ignored. The problem is that you can't see the annotation at the point of use (e.g. in an IDE autocomplete situation). If you have a class in an `impl` package then users are less likely to import them into their code.

What's an example of something that is a limited audience API that can't go in the public API or the SPI?

I opened https://issues.cloudera.org/browse/CDK-878 to discuss this further.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/CSVFileReader.java
rdblue: I opened [CDK-882](https://issues.cloudera.org/browse/CDK-882) for this.

rdblue:
With CDK-882 and CDK-878 filed, I don't think there are any remaining issues before this can be merged.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/CSVFileReader.java
tomwhite: There's still a comment ("avoid divide by 0 errors") which suggests that there are no divide-by-zero errors.

PR #287, opened 2015-01-08
CDK-870: Use blank skin instead of custom empty skin.

rdblue:
The blank-maven-skin is already published and works when the content is
embedded in the jekyll site as well.


joey:
+1, ship it!

PR #288, opened 2015-01-10
CDK-872: Add JSON format

rdblue:
Adds:
* JSON utilities for schema inference and record construction
* JSON format, reader implementation, and InputFormat
* Schema merging util
* CLI commands: json-schema and json-import


joey:
I'm still reviewing this, but I'm curious what are the file format requirements to use this format?

JSON is a record serialization, not a data file format so I assume we have to impose additional constraints on how the records are written to the file.

rdblue:
The Jackson parser currently reads files of concatenated JSON objects that are optionally separated by whitespace. I think we will probably want to also support reading a JSON list as a series of records if one is at the root of the file. For now, I just wanted to get something out and I thought that using what Jackson supports initially would be a good place to start. I'm sure we'll get requests for extending it based on formats that people use, including support for Avro's hinted JSON.

joey:
Thanks. So long as we documented the supported formats, I think we'll be good here. We can add more formats as we need to.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/JsonUtil.java
```diff
@@ -79,4 +117,449 @@ public static JsonNode parse(InputStream in) {
...
+          }
+          break;
+        case FLOAT:
+          if (datum.isFloat() || datum.isInt()) {
+            return schema;
+          }
+          break;
+        case DOUBLE:
+          if (datum.isDouble() || datum.isFloat() ||
+              datum.isLong() || datum.isInt()) {
```
joey: I'm not sure the logic is quite right here. If I have a union of `[double, long]` and in my JSON I have an integral `number`, then I'd expect the resolved type of my Avro schema to be `long`, rather than `double`. From my reading of this code, we'd end up with a `double` because `double` is listed first in the union order. I would think that we need to look for the schema that is the most like the value, rather than the first compatible type.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/JsonUtil.java
```diff
@@ -79,4 +117,449 @@ public static JsonNode parse(InputStream in) {
...
+      this.name = name;
+    }
+
+    public JsonSchemaVisitor useMaps() {
+      this.objectsToRecords = false;
+      return this;
+    }
+
+    @Override
+    public Schema object(ObjectNode _, Map<String, Schema> fields) {
```
joey: We shouldn't use `_` as an identifier as Java may make it a keyword in a future release.

From http://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.27.1:

**It is a compile-time error if a lambda parameter has the name `_` (that is, a single underscore character).**

*The use of the variable name `_` in any context is discouraged. Future versions of the Java programming language may reserve this name as a keyword and/or give it special semantics.*

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/JsonUtil.java
```diff
@@ -79,4 +117,449 @@ public static JsonNode parse(InputStream in) {
...
+          case 1:
+            return Schema.createMap(Iterables.getOnlyElement(fields.values()));
+          default:
+            return Schema.createMap(SchemaUtil.mergeOrUnion(fields.values()));
+        }
+      }
+    }
+
+    @Override
+    public Schema array(ArrayNode _, List<Schema> elementSchemas) {
```
joey: Same note about `_`. I'll assume that two reminders is sufficient :)

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/SchemaUtil.java
```diff
@@ -410,4 +418,404 @@ public static String toString(Object value, Schema schema) {
...
+      case ENUM:
+        if (!Objects.equal(left.getName(), right.getName())) {
+          return null;
+        }
+        Set<String> symbols = Sets.newLinkedHashSet();
+        symbols.addAll(left.getEnumSymbols());
+        symbols.addAll(right.getEnumSymbols());
+        return Schema.createEnum(
+            left.getName(),
+            coalesce(left.getDoc()),
```
joey: Shouldn't we be coalescing the `left.getDoc()` with `right.getDoc()`?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/SchemaUtil.java
```diff
@@ -410,4 +418,404 @@ public static String toString(Object value, Schema schema) {
...
+            coalesce(left.getNamespace()),
```
joey: Same question with namespaces.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/JSONFileReader.java
```diff
@@ -0,0 +1,192 @@
...
+
+  @Override
+  public boolean isOpen() {
+    return (this.state == ReaderWriterState.OPEN);
+  }
+
+  public RecordReader<E, Void> asRecordReader() {
+    Preconditions.checkArgument(incoming instanceof FSDataInputStream,
+        "Cannot use {} in a record reader", incoming.getClass());
+    return new CSVRecordReader();
```
joey: Rename to `JsonRecordReader`.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/JSONFileReader.java
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  public class CSVRecordReader extends RecordReader<E, Void> {
```
joey: Rename to `JsonRecordReader`.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/JSONInputFormat.java
```diff
@@ -0,0 +1,68 @@
...
+    this.descriptor = descriptor;
+  }
+
+  public void setType(Class<E> type) {
+    this.accessor = DataModelUtil.accessor(type, descriptor.getSchema());
+  }
+
+  @Override
+  protected boolean isSplitable(JobContext context, Path filename) {
+    return false;
```
joey: It would be useful to have an option the user could set on if the files would be splittable. For example, I think one of the more common JSON file formats is newline separated JSON objects which would be splittable by newlines.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/Main.java
```diff
@@ -96,9 +98,11 @@
     jc.addCommand("schema", new SchemaCommand(console));
     jc.addCommand("info", new InfoCommand(console));
     jc.addCommand("show", new ShowRecordsCommand(console));
+    jc.addCommand("obj-schema", new ObjectSchemaCommand(console));
     jc.addCommand("csv-schema", new CSVSchemaCommand(console));
     jc.addCommand("csv-import", new CSVImportCommand(console));
-    jc.addCommand("obj-schema", new ObjectSchemaCommand(console));
+    jc.addCommand("json-schema", new JSONSchemaCommand(console));
+    jc.addCommand("json-import", new JSONImportCommand(console));
```
joey: I was talking to @busbey and it'd be nice if we could have a single `import` command and you could pass it an option of JSON, CSV, or an InputFormat. Do you think that would be doable?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/JsonUtil.java
rdblue: I think you're right. [GenericData](https://github.com/apache/avro/blob/trunk/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java) resolves unions by first inspecting the type and seeing if it is an option. This should probably go into this loop only if that strategy doesn't work.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/SchemaUtil.java
rdblue: Yes

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/SchemaUtil.java
rdblue: Yes

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/JSONInputFormat.java
rdblue: My reasoning was the same as for CSV. We can add splitting later if it ever makes sense, but for now it is not very straight-forward (strings can contain newlines).

rdblue: By the way, I think the way to split JSON is to look for symbol pairs that don't occur in valid JSON, like }{. That signals that one record has ended, but isn't separated from the next with a comma so it must be a new root record.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/Main.java
rdblue: Possible, yes. We would probably want to implement the same sub-command structure that is used by Main. But this seems like a lot of work for little gain to me. What is the main benefit of switching from `kite-dataset csv-import ...` to `kite-dataset import csv ...`?

We would still want separate commands so we don't end up with a huge list of options that is the union of CSV, InputFormat, and JSON options. That would be a mess and would be more difficult to keep up to date so users know what options can be applied to each command.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/JSONInputFormat.java
joey: That's why I was saying it should be an option, and not splittable by default. If `}{` is not valid internal to JSON record, then we could use that to be even more robustness. This is even more apropos to my comment about documenting the file format(s) we expect/can handle.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/Main.java
joey: Hrm, you're right that we do need a balance between the number of individual commands we have to expose versus the number of options per command. I'm not sure exactly where that balance is. I'll keep thinking of a better name for informat-import and think more on if there's an elegant way to consolidate the three commands at all.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/JSONInputFormat.java
rdblue: I'm not sure I understand your option suggestion. If we implement splitting, then it is splittable. I agree that there should be an option for turning it off in case of problems. But I don't understand why we would implement an option before we implement splitting. For now, this is always false because we haven't implemented the `}{` strategy or validated that it should work (there are edge cases like ensuring we always start at the beginning of a symbol).

Definitely agree that the supported format should be documented.

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/TestJsonUtil.java
```diff
@@ -0,0 +1,459 @@
...
+
+  @Test
+  public void testSimpleSchemaMerge() throws Exception {
+    String jsonSample = "{\"id\": 1}\n{\"id\": 2}";
+
+    Schema schema = Schema.createRecord("Test", null, null, false);
+    schema.setFields(ImmutableList.of(
+        new Schema.Field("id", Schema.create(Schema.Type.INT), null, null)
+    ));
+
```
tomwhite: SchemaBuilder would make this shorter and more readable.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/JsonUtil.java
```diff
@@ -16,18 +16,56 @@
 
 package org.kitesdk.data.spi;
 
+import com.fasterxml.jackson.core.JsonFactory;
```
tomwhite: Can we shade Jackson please to avoid all the compatibility problems we've seen with it.

tomwhite:
This is a great addition!

As a part of this, I think it would be very useful to add a few lines of documentation explaining how schemas are inferred from JSON documents.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/JSONInputFormat.java
joey: It would be base don the format of the file. I can see at least three common file formats, there might be more:

1. Newline delimited JSON objects without newlines internal to the objects
2. Whitespace delimited JSON objects with embedded newlines possible
3. A list of JSON objects where the top-level JSON element is an array

Format (1) is splittable on newlines. What' I'm proposing is we could add that as an option now as it would hit a good number of use cases. Format (2) is a more general version of format (1) and the `} {` trick might work for both, but as you said we haven't implemented it yet.

Format (3) may never be splittable as it will use commas in between records (`[ {...}, {...}, ...]`). If we want/need to support that format as well then we'll have to be able to configure the file format we're reading.

tomwhite: Splittability is very important for large datasets. I didn't realize that the CSVInputFormat is not splittable at all. Not all CSV is multiline (I would guess most isn't) so we should support splitting on line endings, as well as making it possible to suppress splitting in the multiline case. In fact we should use TextInputFormat since it has logic for handling splits, cross-platform line endings, and for coping with errors caused by long lines.

Similarly, for this change we should support splitting on line endings (Joey's option 1), although perhaps not by default.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/JsonUtil.java
rdblue: I opened [CDK-879](https://issues.cloudera.org/browse/CDK-879) to track this. I'd rather do it in a follow-up issue because there are already components that use Jackson so it isn't a new problem and might run into issues that aren't related to this work.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/Main.java
rdblue: What do you think about separating the top-level commands like this?

* kite-dataset - dataset operations: create, delete, update, show, copy
* kite-import - import for various formats: csv, json, inputformat, etc.
* kite-config - generate configuration: csv-schema, json-schema, flume, partitioning, mapping

Or maybe break kite-config into kite-config and kite-schema?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/JSONInputFormat.java
rdblue: Some JSON is splittable on newlines, but it is more complicated:
* "Whitespace can be inserted between any pair of tokens," so newlines aren't a universal record delimiter
* The parser doesn't yeild whitespace tokens, so implementing a newline rule (for opt-in cases that are safe) requires more work
* For encodings that support multi-byte characters, we need to ensure we can start at the next character to scan for `}{`

I also agree that format 3 should be supported, but as a format option. I'm also not sure that would be splittable at all, though it would be possible to concatenate long list object and split between those by looking for `][`.

> Splittability is very important for large datasets.

I agree, but it isn't necessarily a high priority. I'd like to get some basic support in and then assess the priority of splitting CSV and JSON. Right now we encourage a one-time import and haven't had anyone asking for splitting. We can also wrap dataset directories, but we don't currently expose a convenient way to set that up (a CLI command).

I've opened [CDK-880](https://issues.cloudera.org/browse/CDK-880) to make CSV splittable and [CDK-881](https://issues.cloudera.org/browse/CDK-881) to make JSON splittable. I don't think CSV splits are too hard, but let's follow up on that issue to discuss how to do them properly.

rdblue:
I've updated this for the review comments and pushed a new version (which has passed for the default profile at least). I've also added JSON format docs in https://github.com/kite-sdk/kite-docs/pull/71

I think the remainder of the issues with this have been addressed or are follow-up issues, so this should be ready for another look. Thanks for the reviews!

joey:
+1, I'm fine with the remaining issues being follow-ups.

PR #290, opened 2015-01-20
[CDK-487] Initial pass at a tarfile ingest tool

ianbuss:
Simple tar import tool to import tarballs into a fixed Avro schema with no, gzip and bzip2 compression.


kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
```diff
@@ -0,0 +1,167 @@
...
+import org.kitesdk.data.spi.DatasetRepository;
+import org.slf4j.Logger;
+
+import java.io.IOException;
+import java.util.List;
+
+@Parameters(commandDescription="Import files in tarball into a Dataset")
+public class TarImportCommand extends BaseDatasetCommand {
+
+  private static final List<String> SUPPORTED_TAR_COMPRESSION_TYPES = Lists.newArrayList("none", "gzip", "bzip2");
```
rdblue: Nit: we try to wrap lines at 80 characters unless there are version control reasons not to (e.g., a small addition to a single line is fine)

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
```diff
@@ -0,0 +1,167 @@
...
+        "Tar path does not exist: " + source);
+
+    Preconditions.checkArgument(SUPPORTED_TAR_COMPRESSION_TYPES.contains(compressionType),
+        "Compression type " + compressionType + " is not supported");
+
+    String datasetName = targets.get(1);
+
+    int success = 0;
+
+    DatasetRepository repo = getDatasetRepository();
```
rdblue: We added a load method so you don't have to mess with repositories or URIs. Just [`load(datasetArg)`](https://github.com/kite-sdk/kite/blob/master/kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/CSVImportCommand.java#L127)

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
```diff
@@ -0,0 +1,167 @@
...
+
+    Dataset<GenericRecord> tarImport = null;
+    if (!repo.exists(namespace, datasetName)) {
+      console.info("Creating dataset {} in repo {}", datasetName, repo.getUri());
+      tarImport = repo.create(namespace, datasetName,
+          new DatasetDescriptor.Builder()
+              .schema(new Schema.Parser().parse(
```
rdblue: Could you instead use `schemaUri("resource:tar-import.avsc")`? That way you always use the same schema.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
```diff
@@ -0,0 +1,167 @@
...
+    String datasetName = targets.get(1);
+
+    int success = 0;
+
+    DatasetRepository repo = getDatasetRepository();
+
+    Dataset<GenericRecord> tarImport = null;
+    if (!repo.exists(namespace, datasetName)) {
+      console.info("Creating dataset {} in repo {}", datasetName, repo.getUri());
+      tarImport = repo.create(namespace, datasetName,
```
rdblue: Other import commands require that the dataset already exists. In this case, I think it makes sense to automatically create it because the schema is static. I'm not sure how well that fits with the rest of the tool though. Maybe this would make sense as a its own Kite-based CLI tool?

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
```diff
@@ -0,0 +1,167 @@
...
+                      "    \"name\": \"filecontent\"\n" +
+                      "  }\n" +
+                      "  ]\n" +
+                      "}"
+              ))
+              .build()
+      );
+    } else {
+      console.info("Using existing dataset {} at {}", datasetName, repo.getUri());
+      tarImport = repo.load(namespace, datasetName);
```
rdblue: This needs to check that the expected schema (from the descriptor created above) can work with the loaded dataset.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    DatasetWriter<GenericRecord> writer = tarImport.newWriter();
+
+    // Create a Tar input stream wrapped in appropriate decompressor
+    // Enhancement would be to use native compression libs
+    TarInputStream tis = null;
+    if (compressionType.equals("gzip")) {
+      tis = new TarInputStream(new GzipCompressorInputStream(sourceFS.open(source)));
```
rdblue: Instead of soruceFS.open, we now have an [`open` utility method](https://github.com/kite-sdk/kite/blob/master/kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/BaseCommand.java#L177) as well as a [`output` utility method](https://github.com/kite-sdk/kite/blob/master/kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/BaseCommand.java#L76) that will do the right thing with less logic in the command.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
```diff
@@ -0,0 +1,167 @@
...
+      int count = 0;
+      boolean readError = false;
+      while ((entry = tis.getNextEntry()) != null && !readError) {
+        if (!entry.isDirectory()) {
+          long size = entry.getSize();
+          byte[] buf = new byte[(int) size];
+          int read = 0;
+          int br = 0;
+          // TarInputStream does not always read what you ask for so loop until it does
+          // Note that this requires the entire tar entry (not entire file) to fit into memory
```
rdblue: This seems like an odd note, considering the files that we add are expected to be fairly small. Should this add a maximum file size? Or maybe handle large files by putting them in HDFS directly?

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    return success;
+  }
+
+  @Override
+  public List<String> getExamples() {
+    return Lists.newArrayList(
+            "# Copy the contents of from sample.tar.gz to dataset \"sample\"",
+            "csv-import path/to/sample.tar.gz sample --compression gzip",
```
rdblue: You can remove the "csv-import" from these examples

kite-tools-parent/pom.xml
```diff
@@ -180,6 +180,12 @@
       <scope>compile</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.kamranzafar</groupId>
```
rdblue: Is this a commonly-used tar helper library?

rdblue:
Thanks for building this! It looks like a great addition.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
ianbuss: I'll happily move this is it makes sense for it to be somewhere else. I agree it's not quite like any of the other commands but it does allow me to make easy use of the stuff BaseDatasetCommand.

kite-tools-parent/pom.xml
ianbuss: I've used this one before without any issues but should definitely use existing Apache functionality if possible. I'll play with the stuff that came over from ant into the commons compress lib.

ianbuss:
Thanks for the comments, I'll get on to fixing them as soon as possible.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
ianbuss: Yeah I don't like this either, hence the note. Ideally I could stream into an Avro dataset writer but from what I remember there wasn't an API to do this at the time. I'll check again.

rdblue: I think it probably makes some sense to limit the size. Avro blocks (if I remember correctly) are about 1MB, but records are never split across blocks. That means this will probably stress Avro's assumptions about the data it stores. That should be okay within reason, but I'd like to avoid the situation where the Avro block starts getting within an order of magnitude of the HDFS block size. That would degrade the performance when reading from input splits.

I realize it's a lot more work and would require some custom FS code to build a large file store inside the dataset, so I don't think this is a blocker to get the feature in.

It would be great to get @tomwhite's thoughts on this as well.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
rdblue: Let's leave it here for now. We're considering splitting the kite-dataset command into a couple others since it is growing so large so this can be done at that time.

Any comments @joey or @tomwhite?

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
rdblue: An easy way to do this is to use a Specific class based on the Avro schema definition file (just drop it into `src/main/avro` and make sure the avro plugin is in the pom). That way, Kite will verify that the schema for you.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
```diff
@@ -0,0 +1,167 @@
...
+    Preconditions.checkArgument(SUPPORTED_TAR_COMPRESSION_TYPES.contains(compressionType),
+        "Compression type " + compressionType + " is not supported");
+
+    String datasetName = targets.get(1);
+
+    int success = 0;
+
+    DatasetRepository repo = getDatasetRepository();
+
+    Dataset<GenericRecord> tarImport = null;
```
tomwhite: Nit: you don't need to initialize to null here since the variable is always set. This applies in other places too.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
```diff
@@ -0,0 +1,167 @@
...
+      console.info("Using existing dataset {} at {}", datasetName, repo.getUri());
+      tarImport = repo.load(namespace, datasetName);
+    }
+
+    DatasetWriter<GenericRecord> writer = tarImport.newWriter();
+
+    // Create a Tar input stream wrapped in appropriate decompressor
+    // Enhancement would be to use native compression libs
+    TarInputStream tis = null;
+    if (compressionType.equals("gzip")) {
```
tomwhite: Instead of having the user specify the compression, why not infer from the file extension? .tar -> none, .tar.gz -> gzip, .tar.bz2 -> bzip2

kite-tools-parent/kite-tools/src/test/java/org/kitesdk/cli/commands/TestTarImportCommand.java
```diff
@@ -0,0 +1,188 @@
...
+            Random random = new Random(1);
+            for (int i = 0; i < NUM_TEST_FILES; ++i) {
+                // Create random file and data
+                int fNameLength = random.nextInt(MAX_FILENAME_LENGTH);
+                int fContentLength = random.nextInt(MAX_FILECONTENT_LENGTH);
+                String fName = RandomStringUtils.randomAlphanumeric(fNameLength);
+                String fContent = RandomStringUtils.randomAscii(fContentLength);
+
+                // Write the file to tarball
+                TestTarImportCommand.writeToTarFile(tosNoCompression, TAR_TEST_ROOT_PREFIX + "/" + fName, fContent);
```
tomwhite: Nit: no need to qualify with class.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
tomwhite: Avro records are not split across blocks, but blocks have a minimum size, not a fixed size. So things will work fine even if you have very large records - it's just that splitting is less efficient when reading (as Ryan notes). As the expected use case is importing tar files from local file systems, I don't think we need to do anything particularly complicated here. Maybe just warn if a record is bigger than an HDFS block?

As for streaming the bytes - there's no mechanism for doing that since you have to build the GenericRecord in memory. I could imagine a specialized version of GenericRecord that uses an InputStream instead of a byte array for an Avro bytes type, but again I don't think we need to do this given the expected use case (where individual records are not greater than an HDFS block).

Also, you might use IOUtils#readFully from Commons Compress that will read a given number of bytes into a byte array: https://commons.apache.org/proper/commons-compress/javadocs/api-1.9/index.html?org/apache/commons/compress/utils/IOUtils.html

kite-tools-parent/kite-tools/src/test/java/org/kitesdk/cli/commands/TestTarImportCommand.java
```diff
@@ -0,0 +1,188 @@
...
+
+    @Test
+    public void testTarNoCompressionImportCommand() throws IOException {
+        command.targets = Lists.newArrayList(TAR_TEST_FILE, TEST_DATASET_NAME);
+        command.compressionType = "none";
+        command.run();
+        verify(console).info("Added {} records to \"{}\"",
+                NUM_TEST_FILES,
+                "repo:" + cwd + "/" + TEST_DATASET_DIR + "::" + TEST_DATASET_NAME);
+        //verifyNoMoreInteractions(console);
```
tomwhite: Any reason these are commented out?

kite-tools-parent/pom.xml
tomwhite: +1 to using Apache Commons Compress since it's already a dependency.

tomwhite:
This is a great feature to add. Thanks for working on it Ian!

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
tomwhite: Also, for the enhancement you could use Hadoop's CompressionCodec#createInputStream to use native libs. Not needed initially though.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
```diff
@@ -0,0 +1,177 @@
...
+    String source = targets.get(0);
+    String datasetUri = targets.get(1);
+
+    long blockSize = getConf().getLong("dfs.blocksize", DEFAULT_BLOCK_SIZE);
+
+    int success = 0;
+
+    View<TarFileEntry> targetDataset;
+    if (Datasets.exists(datasetUri)) {
+      console.info("Using existing dataset: {}", datasetUri);
```
rdblue: Nit: I think this one should be debug because this is what the user instructed the command to do. The unexpected case is when the dataset must be created.

kite-tools-parent/kite-tools/src/test/java/org/kitesdk/cli/commands/TestTarImportCommand.java
```diff
@@ -0,0 +1,258 @@
...
+    if (null != content) {
+      byte[] buf = content.getBytes();
+      tos.write(buf, 0, content.length());
+      tos.flush();
+    }
+    tos.closeArchiveEntry();
+  }
+
+  @Test
+  public void testTarNoCompressionImportCreateDataset() throws Exception {
```
rdblue: Should this verify the create message printed to the console?

kite-tools-parent/kite-tools/src/test/java/org/kitesdk/cli/commands/TestTarImportCommand.java
```diff
@@ -0,0 +1,258 @@
...
+    Path testData = new Path(TEST_DATA_DIR);
+    FileSystem testFS = testData.getFileSystem(new Configuration());
+
+    datasetUri = "dataset:file:" + System.getProperty(
+        "user.dir") + "/" + TEST_DATASET_DIR + "/" + TEST_DATASET_NAME;
+
+    TarArchiveOutputStream tosNoCompression = null;
+    TarArchiveOutputStream tosGzipCompression = null;
+    TarArchiveOutputStream tosBzip2Compression = null;
+    XZCompressorOutputStream xzos = null;
```
rdblue: This doesn't appear to be used

rdblue:
@ianbuss great job! Only minor stuff left, we should be able to merge this really soon.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/TarImportCommand.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.List;
+
+@Parameters(commandDescription = "Import files in tarball into a Dataset")
+public class TarImportCommand extends BaseDatasetCommand {
+
+  protected enum CompressionType {
+    NONE,
+    GZIP,
+    BZIP2,
+    XZ
```
tomwhite: Nit: not used

kite-tools-parent/kite-tools/src/test/resources/test-schemas/tar-import.avsc
```diff
@@ -0,0 +1,14 @@
+{
+ "type": "record",
+ "name": "tarimport",
+ "fields": [
```
tomwhite: This file is a duplicate of the one in main. You can reference the schema from TarFileEntry instead.

tomwhite:
I tried this out manually and it worked well. Just a couple of minor comments to address.

rdblue:
The test failure is related to me breaking master, not this commit.

rdblue:
Merged! Thanks Ian!

ianbuss:
Cool! Thanks

PR #292, opened 2015-01-24
CDK-888: Improve error handling for CSV and JSON.

rdblue:
This adds the field name to help users narrow down what data or schema
doesn't match.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetRecordException.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetRecordException.java
@@ -33,4 +33,24 @@ public class DatasetRecordException extends DatasetException {
   public DatasetRecordException(String message, Throwable t) {
     super(message, t);
   }
+
+  /**
+   * Precondition-style validation that throws a {@link DatasetRecordException}.
+   *
+   * @param isValid
+   *          {@code true} if valid, {@code false} if an exception should be
+   *          thrown
+   * @param message
+   *          A String message for the exception.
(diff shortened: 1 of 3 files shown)


kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetRecordException.java
```diff
@@ -33,4 +33,24 @@ public DatasetRecordException(String message) {
...
+   *          A String message for the exception.
+   */
+  public static void check(boolean isValid, String message, Object... args) {
+    if (!isValid) {
+      String[] argStrings = new String[args.length];
+      for (int i = 0; i < args.length; i += 1) {
+        argStrings[i] = String.valueOf(args[i]);
+      }
+      throw new DatasetRecordException(
+          String.format(String.valueOf(message), (Object[]) argStrings));
```
tomwhite: Nit: `String.valueOf(message)` is redundant

tomwhite:
One nit, otherwise +1

kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetRecordException.java
rdblue: This handles the case where message is null. Instead of the empty string or a lurking NPE from String.format that suppresses a real exception, you get "null".

tomwhite: Makes sense. +1

PR #296, opened 2015-01-26
CDK-879. Remove use of Jackson 1 API (in org.codehaus.jackson package).

tomwhite:
This removes all of the Jackson 1 usages, except in kite-data-hbase's AvroUtils, where we are using an Avro API that requires it.


rdblue:
+1

rdblue:
Should we look at adding a copy constructor to Field in Avro to avoid the need for a Jackson-dependent method in Kite?

tomwhite:
We could, although that wouldn't be needed with the work to make Schema (and Field) immutable. See https://issues.apache.org/jira/browse/AVRO-1608.

tomwhite:
Merged in https://github.com/kite-sdk/kite/commit/3598bc67ecb7665c30c2c8c1d9892d0332be53ec

PR #297, opened 2015-01-27
Add SchemaUtil.digest.

rdblue:
WIP.

This will help keep track of Schemas by providing a reliable hashing function. This uses the Schema structure directly to calculate a digest, so the digest doesn't change if formatting does. Open questions:
* This currently ignores default values: should defaults be optional or included?
* The digest depends on union, enum, and record ordering, but should `h([null, long]) == h([long, null])`?
* Is SHA-1 a reasonable algorithm? Is there a faster alternative we should use?


tomwhite:
Avro defines a canonical form for schemas and schema fingerprints (http://avro.apache.org/docs/1.7.7/spec.html#Parsing+Canonical+Form+for+Schemas), which should be suitable for this purpose. See https://avro.apache.org/docs/1.7.7/api/java/org/apache/avro/SchemaNormalization.html.

PR #302, opened 2015-02-04
CDK-899: Update Datasets.java Javadoc entries.

DennisDawson:


kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -31,25 +31,52 @@
...
- *
+ * Methods for working with {@link Dataset} instances.
+ * <p>
+ * <b>URIs</b><br />
+ * All methods require a URI that identifies a dataset, view, or
+ * repository. The URI must begin with the scheme {@code dataset:}, 
+ * {@code view:}, or {@code repo:}. The remainder of the URI is 
+ * implementation specific, depending on the dataset scheme.
+ * <p>
+ * For example, the URI {@code dataset:hive:/movies/ratings}
```
rdblue: This example shouldn't have the '/' before movies.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -31,25 +31,52 @@
...
+ * All methods require a URI that identifies a dataset, view, or
+ * repository. The URI must begin with the scheme {@code dataset:}, 
+ * {@code view:}, or {@code repo:}. The remainder of the URI is 
+ * implementation specific, depending on the dataset scheme.
+ * <p>
+ * For example, the URI {@code dataset:hive:/movies/ratings} 
+ * references a Hive dataset named <i>ratings</i> in the
+ * <i>movies</i> namespace.
+ * <p>
+ * The URI {@code view:hdfs:/user/myfiles/movies/ratings?year=2015&month=3}
```
rdblue: This uses "myfiles" where you would normally see a user name, implying that the user is actually named "myfiles". I think it would be better to use a generic user name there. I've been using "/user/me/movies/ratings".

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -31,25 +31,52 @@
...
+ * references a view of an HDFS dataset named <i>ratings</i> in the
+ * <i>movies</i> namespace in the {@code /user/myfiles} path. The view 
+ * is filtered to include records from only March, 2015.
+ * <p>
+ * See <a href="http://kitesdk.org/docs/current/URIs.html">Dataset and View
```
rdblue: What about changing the link text to "available URIs"? I think it could be more clear why someone should follow this link.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -31,25 +31,52 @@
...
+ * URIs</a>.
+ * <p>
+ * <b>Entities</b><br />
+ * <i>Entities</i> are analagous to <i>records</i> in database terminology.
```
rdblue: I thought we discussed moving away from _entities_ in favor of _records_. Should this be updated?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -31,25 +31,52 @@
...
+ * The term is used in the API to emphasize that an entity can include not
+ * only primitive objects, but also complex objects such as hash maps. 
+ *
```
rdblue: Since this is calling out the use of record classes and URIs, what about a short note on the DatasetDescriptor and a link to its builder?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -31,25 +31,52 @@
...
+   * metadata, and verifies that you can communicate with its services.
    * <p>
-   * If the URI is a dataset URI, the unfiltered Dataset will be returned.
+   * URIs must begin with {@code dataset:} or {@code view:}. The remainder of
+   * the URI is implementation specfic, depending on the dataset scheme.
    * <p>
-   * URI formats are defined by {@code Dataset} implementations, but must begin
-   * with "dataset:" or "view:".
+   * If you use a dataset URI, {@code load} returns the unfiltered dataset.
+   * If you use a view URI, {@code load} returns a subset of the data, based on
```
rdblue: This doesn't return a subset of the data. Instead, it returns a View configured to read a subset of the data. Does that distinction make sense?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -31,25 +31,52 @@
...
+   * If you use a view URI, {@code load} returns a subset of the data, based on 
+   * your filter criteria.
    *
-   * @param uri a {@code Dataset} or {@code View} URI.
-   * @param type the Java type of the entities in the dataset
-   * @param <E> The type of entities stored in the {@code Dataset}.
-   * @param <V> The type of {@code View} expected.
-   * @return a {@code View} for the given URI.
+   * @param uri a {@code Dataset} or {@code View} URI
+   * @param type a Java object that represents an entity in the dataset
```
rdblue: Instead of changing "type" to "object", it would be better to say "class". This is the class that should be returned by readers created by the dataset or view loaded by this method.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -31,25 +31,52 @@
...
+   * @param <E> the type of entities stored in the {@code Dataset}
+   * @param <V> the type of {@code View} expected
+   * @return a {@code View} for the given URI
```
rdblue: This should have `@throws` tags for all of the exceptions that could be thrown by `DatasetRepository`, which is the backing class.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -31,25 +31,52 @@
...
+   * your filter criteria.
    *
-   * @param uri a {@code Dataset} or {@code View} URI.
-   * @param type the Java type of the entities in the dataset
-   * @param <E> The type of entities stored in the {@code Dataset}.
-   * @param <V> The type of {@code View} expected.
-   * @return a {@code View} for the given URI.
+   * @param uri a {@code Dataset} or {@code View} URI
+   * @param type a Java object that represents an entity in the dataset
+   * @param <E> the type of entities stored in the {@code Dataset}
```
rdblue: This is the type that is expected when reading, rather than the type that is stored. The type that is stored is represented by the dataset's Schema. This is an actual class that will be used to hold records described by that Schema. For example, you can use a `GenericData.Record` object to hold _any_ schema's data. You might also use a specific object that has been generated from a Schema.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -31,25 +31,52 @@
...
+ * The term is used in the API to emphasize that an entity can include not
+ * only primitive objects, but also complex objects such as hash maps. 
+ *  
  * @since 0.8.0
  */
 public class Datasets {
 
   /**
    * Load a {@link Dataset} or {@link View} for the given {@link URI}.
+   * {@code load} verifies that the dataset exists, retrieves the dataset's
```
rdblue: It is sufficient to say that it loads the dataset and retrieves its metadata. I wouldn't say that this verifies that a dataset exists or that it verifies that you can communicate with its services. In the first case, it doesn't specifically check. It just throws an exception if there is no dataset and it can't proceed.

For the part about communicating with its services, we don't really think of a dataset as something that has services. I think you mean that it verifies the communication with any services needed to access the data, but that isn't strictly true. We could load information from the Hive MetaStore without necessarily verifying that we can communicate with HDFS.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -257,12 +310,18 @@
   /**
    * Update a {@link Dataset} for the given dataset or view URI.
    * <p>
-   * URI formats are defined by {@code Dataset} implementations, but must begin
-   * with "dataset:" or "view:".
+   * You can add, remove, or change the datatype of columns in your dataset,
```
rdblue: I think this should be reworded. It isn't true that you can remove the datatype, but I don't think that is what you were trying to say.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -257,12 +310,18 @@
...
+   * provided you don't attempt a change that would corrupt the data. Kite
+   * follows the Avro guidelines for updating schemas. See 
+   * <a href="http://kitesdk.org/docs/current/Schema-Evolution.html">Schema
```
rdblue: Can you use `@link` here instead of HTML?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -225,14 +272,20 @@
   /**
    * Update a {@link Dataset} for the given dataset or view URI.
    * <p>
-   * URI formats are defined by {@code Dataset} implementations, but must begin
-   * with "dataset:" or "view:".
+   * You can add, remove, or change the datatype of columns in your dataset,
```
rdblue: I think ["datatype" should be "data type"](https://books.google.com/ngrams/graph?content=data+type%2C+datatype&year_start=1800&year_end=2000&corpus=15&smoothing=3&share=&direct_url=t1%3B%2Cdata%20type%3B%2Cc0%3B.t1%3B%2Cdatatype%3B%2Cc0)

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -327,26 +397,27 @@ public static boolean delete(URI uri) {
...
    * <p>
-   * URI formats are defined by {@code Dataset} implementations, but must begin
-   * with "dataset:".
+   * URIs must begin with {@code dataset:}. The remainder of
+   * the URI is implementation specfic, depending on the dataset scheme.
    *
-   * @param uri a {@code Dataset} URI.
-   * @return {@code true} if the dataset exists, {@code false} otherwise
+   * @param uri a {@code Dataset} URI
+   * @return {@code true} if the dataset exists, or {@code false}
```
rdblue: I think "otherwise" is more clear. Would "otherwise false" work?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -364,23 +435,25 @@ public static boolean exists(URI uri) {
...
 
   /**
-   * List the {@link Dataset} URIs in the repository identified by the URI
+   * List the {@link Dataset} URIs in the repository identified by the URI.
    * <p>
    * URI formats are defined by {@code Dataset} implementations. The repository
-   * URIs passed to this method must begin with "repo:".
+   * URIs you pass to this method must begin with {@code repo:}. For example, to 
+   * list the {@code Dataset} URIs for the Hive repository {@code /default}, 
+   * provide the URI {@code repo:hive:/default}.
```
rdblue: Hive is the repository, which contains databases (namespaces), which contain tables (datasets). The repository URI for Hive is "repo:hive".

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Removed the forward slash.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Changed to:

 * The URI {@code view:hdfs:/user/me/movies/ratings?year=2015&month=3}
 * references a view of an HDFS dataset named <i>ratings</i> in the
 * <i>movies</i> namespace in the {@code /user/me} path. The view 
 * is filtered to include records from only March, 2015.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: That is the title of the linked document. I don't think this is ambiguous, in context.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: _Entities_ is used in the API. If the term is going to be removed from the API, then we can use _records_ and omit this note.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
rdblue: I just noticed also that the two examples are similar but don't match. Would it be better to use the same dataset to demonstrate the view URI?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
rdblue: Okay, that sounds good for now. I don't really want to remove all the API references right now anyway.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: That is probably a better candidate for packages.html. The URIs section is to give more context for the descriptions of URIs, which appears in every method description. The DatasetDescriptor is linked in every method that uses it. The notes in the header are to clear up what might not be immediately apparent in the short method descriptions.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Changed it. Also saying subset of the dataset, rather than the data, to make it parallel. I made the change in all 4 flavors of the load method.
***
   * If you use a dataset URI, {@code load} returns the unfiltered dataset.
   * If you use a view URI, {@code load} returns a {@code View} configured to read 
   * a subset of the dataset.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Changed _object_ to _class_ in all methods.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: I don't think so. I can't find any syntax that allows it. Here's a link to the Javadoc spec.

http://docs.oracle.com/javase/1.5.0/docs/tooldocs/windows/javadoc.html#{@link}

@see doesn't help, either, because it creates a See Also: section. I want to keep the link under the URIs heading.

I spent a considerable amount of time chasing this down: if you can show a syntax example, please do.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Change 4 instances of datatype to data type.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: I just removed the "false" clause for all four instances. These are booleans, and "false" result is obvious.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Changed to:

   * URI formats are defined by {@code Dataset} implementations. The repository
   * URIs you pass to this method must begin with {@code repo:}. For example, to 
   * list the {@code Dataset} URIs for the Hive repository, provide the URI
   * {@code repo:hive}.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
rdblue: I like for the wording to be more clear. To avoid the problem another way, we can use "{@code true} if and only if the dataset exists."

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Copied the throws tags from DataetRepository for each of the corresponding methods.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Reworded to 

   * You can add columns, remove columns, or change the data type of columns 
   * in your dataset, provided you don't attempt a change that would corrupt 
   * the data.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Put everything back to "false otherwise," which is conventional. (I looked at java.util.String, for instance, and it's used there. When in Rome.)

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Can you please provide the short description I should use to replace the existing description? I don't know how to translate the description you've given into a brief description for the Javadoc. Something like this?

"the type of entities read from the dataset"

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: That description comes verbatim from API-Overview.html#load. Should I file a JIRA to correct this there?

Changed the first paragraph of all Load descriptions to the following:

   * Load a {@link Dataset} or {@link View} for the given {@link URI} and
   * retrieve its metadata.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
rdblue: Works for me.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
rdblue: I think this will work: "the type used readers and writers from this dataset".

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
rdblue: Yes, please file an issue to fix it.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -31,25 +31,52 @@
...
 public class Datasets {
 
   /**
    * Load a {@link Dataset} or {@link View} for the given {@link URI}.
+   * {@code load} verifies that the dataset exists, retrieves the dataset's
+   * metadata, and verifies that you can communicate with its services.
    * <p>
-   * If the URI is a dataset URI, the unfiltered Dataset will be returned.
+   * URIs must begin with {@code dataset:} or {@code view:}. The remainder of
+   * the URI is implementation specfic, depending on the dataset scheme.
```
tomwhite: Typo in "specific". Also in other places in this PR.

DennisDawson: Ouch! Fixed.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: I purposely used a different scheme to show more than one in action. Using the same one would be important if what we're trying to demonstrate that `view:` is substantially different than `dataset:`. What I'm trying to demonstrate is that they are substantially the same, and work in similar fashion with all schemes.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Replaced all instances of the <E> param description with:

   * @param <E> the type used by readers from and writers to the {@code Dataset}

rdblue: Sorry my wording wasn't clear. My version used _from_ to mean that a reader or writer was produced by the dataset. As in, "the writer from the dataset's newWriter method". Could you please revise this?

DennisDawson: Per discussion, changed all instances to:

* @param <E> the type used for readers and writers created by this {@code Dataset}

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Add JIRA issue CDK-903 - Incorrect description of the `load` command in Kite-API

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Replaced references to `name` (from DatasetRepository) with `URI` (for Datasets) in descriptions.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -31,25 +31,53 @@
 import org.kitesdk.data.spi.Registration;
 
 /**
- * <p>Convenience methods for working with {@link Dataset} instances.</p>
- *
+ * Methods for working with {@link Dataset} instances.
+ * <p>
+ * <b>URIs</b><br /><br />
```
rdblue: Why use two breaks instead of a new paragraph?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
rdblue: I'd prefer to change only one thing at a time for examples. If you want to show more than one URI pattern (Hive and HDFS), could you add another example?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
rdblue: I think this still needs more. Ideally, the reader will always have enough information to know whether to click a link. Please fix this to something like "See the available dataset and view URI patterns".

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -31,25 +31,53 @@
...
+ * only primitive objects, but also complex objects such as hash maps. 
+ * 
  * @since 0.8.0
  */
 public class Datasets {
 
   /**
-   * Load a {@link Dataset} or {@link View} for the given {@link URI}.
+   * Load a {@link Dataset} or {@link View} for the given {@link URI} and
+   * retrieve its metadata.
```
rdblue: What does load mean if not to retrieve the dataset's metadata? I don't think the addition is needed.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -31,25 +31,53 @@
...
-   * @param type the Java type of the entities in the dataset
-   * @param <E> The type of entities stored in the {@code Dataset}.
-   * @param <V> The type of {@code View} expected.
-   * @return a {@code View} for the given URI.
+   * @param uri a {@code Dataset} or {@code View} URI
+   * @param type a Java class that represents an entity in the dataset
+   * @param <E> the type used for readers and writers created by this {@code Dataset}
+   * @param <V> the type of {@code View} expected
+   * @return a {@code View} for the given URI
+   * @throws DatasetNotFoundException   if there is no dataset for the given URI}
```
rdblue: Typo: `}` at the end

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -31,25 +31,53 @@
...
+   * @throws DatasetRepositoryException
```
rdblue: This exception no longer exists and references to it can be removed.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -132,16 +175,33 @@
...
+   * @param <V> the type of {@code Dataset} or {@code View} expected
+   * @return a newly created {@code Dataset} responsible for the given URI
+   * @throws IllegalArgumentException   if {@code URI} or {@code descriptor}
+   *                                    is {@code null}
+   * @throws DatasetExistsException     if a {@code Dataset} for the given URI
+   *                                    already exists
+   * @throws ConcurrentSchemaModificationException
+   *                                    if the {@code Dataset}
+   *                                    schema is updated
+   *                                    concurrently
```
rdblue: This formatting leaves a lot of white space and is hard to read. Best practice for this situation is to use one line if the phrase fits on one line, or a constant number of spaces to indent on the following lines for a paragraph.

```
@throws ValidationException if the schema and strategy are incompatible
@throws ConcurrentSchemaModificationException
          If the {@code Dataset} schema is updated while before this operation
          completes and invalidates the requested changes.
```

It is also okay to use the newline and indented text at all times. I recommend 10 spaces for the indent from where `@throws` is placed on the line.

rdblue: I don't think we need `ConcurrentSchemaModificationException` here. It should not be thrown for create because we should throw `DatasetExistsException` instead.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -132,16 +175,33 @@
 
   /**
    * Create a {@link Dataset} for the given dataset or view URI.
+   * {@code create} returns an empty dataset. You can use 
+   * {@link DatasetWriter} to populate your dataset.
    * <p>
-   * URI formats are defined by {@code Dataset} implementations, but must begin
-   * with "dataset:" or "view:".
+   * URIs must begin with {@code dataset:} or {@code view:}. The remainder of
+   * the URI is implementation specific, depending on the dataset scheme.
```
rdblue: If the URI is a view URI, this method creates the underlying dataset and returns a view of it.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -79,51 +107,66 @@
...
-   * @return a {@code View} for the given URI.
-   */
+   * If you use a dataset URI, {@code load} returns the unfiltered dataset.
+   * If you use a view URI, {@code load} returns a {@code View} configured to read 
+   * a subset of the dataset.
+   * 
+   * @param uri a {@code Dataset} or {@code View} URI
+   * @param <V> the type of {@code View} expected
+   * @return a {@code View} for the given URI
+   * @throws DatasetNotFoundException   if there is no dataset for the given URI
```
rdblue: `load` methods will also throw NPE if any of the arguments are null.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -132,16 +175,33 @@
...
-   * @param <E> The type of entities stored in the {@code Dataset}.
-   * @param <V> The type of {@code Dataset} or {@code View} expected.
-   * @return a newly created {@code Dataset} responsible for the given URI.
-   */
+   * @param uri a {@code Dataset} or {@code View} URI
+   * @param type a Java class that represents an entity in the dataset
+   * @param <E> the type used for readers and writers created by this {@code Dataset}
+   * @param <V> the type of {@code Dataset} or {@code View} expected
+   * @return a newly created {@code Dataset} responsible for the given URI
+   * @throws IllegalArgumentException   if {@code URI} or {@code descriptor}
```
rdblue: This should be `NullPointerException`

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -132,16 +175,33 @@
 
   /**
    * Create a {@link Dataset} for the given dataset or view URI.
+   * {@code create} returns an empty dataset. You can use 
+   * {@link DatasetWriter} to populate your dataset.
```
rdblue: Best practice is to use a short sentence to explain what the method does, then start a new paragraph for additional info.

I don't know that we need to link to `DatasetWriter`. What is the value?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -225,14 +336,38 @@
...
-   * @param uri a {@code Dataset} or {@code View} URI.
-   * @param type the Java type of the entities in the dataset
-   * @param <E> The type of entities stored in the {@code Dataset}.
-   * @param <D> The type of {@code Dataset} expected.
-   * @return a newly created {@code Dataset} responsible for the given URI.
+   * @param uri a {@code Dataset} or {@code View} URI
+   * @param type a Java class that represents an entity in the dataset
+   * @param <E> the type used for readers and writers created by this {@code Dataset}
+   * @param <D> the type of {@code Dataset} expected
+   * @return a newly created {@code Dataset} responsible for the given URI
```
rdblue: This shouldn't say it returns a newly created dataset. It should state that it returns a dataset for the given URI.

rdblue: Should also note that views are not returned by this method, even if a View URI is passed in.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -225,14 +336,38 @@
   /**
    * Update a {@link Dataset} for the given dataset or view URI.
    * <p>
-   * URI formats are defined by {@code Dataset} implementations, but must begin
-   * with "dataset:" or "view:".
+   * You can add columns, remove columns, or change the data type of columns 
+   * in your dataset, provided you don't attempt a change that would corrupt 
+   * the data. Kite follows the guidelines in the Avro schema. See
```
rdblue: Rather than "provided you don't attempt a change that would corrupt the data", this should be "provided you don't attempt a change that is incompatible with written data". This avoids suggesting that data can be corrupted.

The sentence "Kite follows the guidelines in the Avro schema" isn't correct. An Avro schema doesn't contain guidelines, but Avro defines rules for compatible schema evolution.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -225,14 +336,38 @@
...
    * Update a {@link Dataset} for the given dataset or view URI.
    * <p>
-   * URI formats are defined by {@code Dataset} implementations, but must begin
-   * with "dataset:" or "view:".
+   * You can add columns, remove columns, or change the data type of columns 
+   * in your dataset, provided you don't attempt a change that would corrupt 
+   * the data. Kite follows the guidelines in the Avro schema. See 
+   * <a href="http://kitesdk.org/docs/current/Schema-Evolution.html">Schema
+   * Evolution</a>.
+   * <p>
```
rdblue: This method updates the dataset's descriptor, not just the dataset's schema. This is how you would add or change properties.

This should link to the DatasetDescriptor.Builder's copy constructor. That allows you to start building a descriptor from an  existing one, which is the recommended way to build descriptors to pass into this method.

Also note that the dataset's format and partitioning can't be changed.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -225,14 +336,38 @@
...
-   * @param type the Java type of the entities in the dataset
-   * @param <E> The type of entities stored in the {@code Dataset}.
-   * @param <D> The type of {@code Dataset} expected.
-   * @return a newly created {@code Dataset} responsible for the given URI.
+   * @param uri a {@code Dataset} or {@code View} URI
+   * @param type a Java class that represents an entity in the dataset
+   * @param <E> the type used for readers and writers created by this {@code Dataset}
+   * @param <D> the type of {@code Dataset} expected
+   * @return a newly created {@code Dataset} responsible for the given URI
+   * @throws IllegalArgumentException      if {@code URI} is null
```
rdblue: This is `NullPointerException` too. Looks like the javadoc you pulled this from wasn't correct; sorry about that!

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -303,13 +510,19 @@
   }
 
   /**
-   * Delete a {@link Dataset} identified by the given dataset URI.
+   * Delete a {@link Dataset} identified by the given dataset URI. Kite takes care
+   * of any housekeeping, including deleting data and any metadata stored separately.
```
rdblue: New paragraph here, too.

rdblue: When called with a dataset URI, both data and metadata will be deleted. After this method is called, the dataset will no longer exist, unless an exception is thrown.

When called with a view URI, data in that matches the view will be deleted. The dataset's metadata is not changed. This may throw `UnsupportedOperationException` if the delete requires additional work. See the [`View#deleteAll docs`](http://kitesdk.org/docs/0.17.1/apidocs/org/kitesdk/data/View.html#deleteAll()) for more information that should be covered here.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -364,26 +585,31 @@ public static boolean exists(URI uri) {
...
    * <p>
-   * URI formats are defined by {@code Dataset} implementations, but must begin
-   * with "dataset:".
+   * URIs must begin with {@code dataset:}. The remainder of
+   * the URI is implementation specific, depending on the dataset scheme.
    *
-   * @param uri a {@code Dataset} URI string.
+   * @param uri a {@code Dataset} URI string
    * @return {@code true} if the dataset exists, {@code false} otherwise
+   * @return a {@link Collection} of Dataset names ({@link String}s)
```
rdblue: Minor: avoid breaking words with javadoc tags. Instead you can use `{@link String strings}`.

http://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html#link

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
rdblue: Would it be more appropriate to use a h2?

DennisDawson: h2 leaves too much space (in fact, this evolved from h2 to one br to two br's). <p> is better, but two br's don't affect the output and should not block publication.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Your earlier comment was "It is sufficient to say that it loads the dataset and retrieves its metadata."  I've removed "and retrieve its metadata" from 4 instances.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Removed right parenthesis from the end of the line.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Removed all references to DatasetRepositoryException.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: This was the existing formatting. It doesn't affect the output. If it is a standard, then it needs to be imposed when the class is added to the project. After 1.0, we can create a set of guidelines and address this sort of cosmetic change.

Removed ConcurrentSchemaModificationException.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Added this note for the 4 instances of create().

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Added NullPointerException to 4 instances of load().

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Changed IllegalArgumentException to NullPointerException for all methods that have an IllegalArgumentException.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: This best practice is not followed in our current Javadocs, nor in Javadocs from Google, Oracle (Sun), VMWare, or others I checked. We can consider making this part of the standard after 1.0.

Removed the link to DatasetWriter.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Changed paragraph in all 4 update instances to:

   * You can add columns, remove columns, or change the data type of columns 
   * in your dataset, provided you don't attempt a change that is incompatible
   * with written data. Avro defines rules for compatible schema evolution. See 
   * <a href="http://kitesdk.org/docs/current/Schema-Evolution.html">Schema
   * Evolution</a>.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
DennisDawson: Changed IllegalArgumentException to NullPointerException throughout.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -96,34 +128,43 @@
   }
 
   /**
-   * Load a {@link Dataset} or {@link View} for the given URI string.
+   * Load a {@link Dataset} or {@link View} for the given {@link URI}.
```
rdblue: In this case, the URI is a String, so the previous version was correct.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -131,17 +172,32 @@
   }
 
   /**
-   * Create a {@link Dataset} for the given dataset or view URI.
+   * Create a {@link Dataset} for the given dataset or view URI string.
```
rdblue: In this case, the URI must be an instance of java.net.URI, not String, so the addition of "string" should be reverted.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -225,14 +324,49 @@
...
+   * with written data. Avro defines rules for compatible schema evolution. See 
+   * <a href="http://kitesdk.org/docs/current/Schema-Evolution.html">Schema
+   * Evolution</a>.
+   * <p>
+   * This method updates the dataset descriptor, so you can also add
+   * or change properties.
+   * <p>
+   * The recommended way to update a dataset descriptor is to build it
+   * based on an existing descriptor. Use
+   * {@link DatasetDescriptor.Builder(DatasetDescriptor descriptor)} to
```
rdblue: Link tags use the type, but not the argument name. This should be `{@link DatasetDescriptor.Builder(DatasetDescriptor)}`

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -304,12 +541,30 @@
 
   /**
    * Delete a {@link Dataset} identified by the given dataset URI.
```
rdblue: This isn't quite correct because view URIs are supported. It should be "Delete a Dataset or View identified by the given dataset URI."

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -304,12 +541,30 @@
...
+  * <p>
+   * When you call this method using a dataset URI, both data and metadata are
+   * deleted. After you call this method, the dataset no longer exists, unless
+   * an exception is thrown.
+   * <p>
+   * When you call this method using a view URI, data in that view is deleted.
+   * The dataset's metadata is not changed. This can throw an
```
rdblue: This should state that the dataset's metadata is not deleted, and partition metadata is updated to reflect deleted data.

rdblue:
Merged in https://github.com/kite-sdk/kite/commit/330f1bdb956145e26d8c6cb93d2952c97e3e52aa.

PR #303, opened 2015-02-05
CDK-898. Importing a large local CSV file causes out of memory error.

tomwhite:
Always use MRPipeline.

### DIFF
--- a/kite-tools-parent/kite-tools/src/main/java/org/kitesdk/tools/TransformTask.java
+++ b/kite-tools-parent/kite-tools/src/main/java/org/kitesdk/tools/TransformTask.java
@@ -96,63 +96,32 @@ public class TransformTask<S, T> extends Configured {
   }
 
   public PipelineResult run() throws IOException {
-    boolean runInParallel = true;
-    if (isLocal(from.getDataset()) || isLocal(to.getDataset())) {
-      runInParallel = false;
-    }
-
     PType<T> toPType = ptype(to);
     MapFn<T, T> validate = new CheckEntityClass<T>(to.getType());
 
-    if (runInParallel) {
-      TaskUtil.configure(getConf())
-          .addJarPathForClass(HiveConf.class);
+    TaskUtil.configure(getConf())
+        .addJarPathForClass(HiveConf.class);
 
-      Pipeline pipeline = new MRPipeline(getClass(), getConf());
+    Pipeline pipeline = new MRPipeline(getClass(), getConf());
 
-      PCollection<T> collection = pipeline.read(CrunchDatasets.asSource(from))
-          .parallelDo(transform, toPType).parallelDo(validate, toPType);
+    PCollection<T> collection = pipeline.read(CrunchDatasets.asSource(from))
+        .parallelDo(transform, toPType).parallelDo(validate, toPType);
 
-      if (compact) {
-        // the transform must be run before partitioning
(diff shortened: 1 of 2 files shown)


rdblue:
This fix always starts an MR job? That seems overkill for most examples.

How does the read from the local FS work? Is it handled by Crunch?

What about detecting large files and basing the MR vs local logic on that?

tomwhite:
We need to use the local job runner and the MRPipeline if either the input or output dataset is a local file. We could also do an optimization to use the MemPipeline if the input dataset is less than a fixed threshold (we can check using SizeAccessor).

rdblue:
Okay, using local job runner makes more sense. I think this needs to include that change so we don't have a regression. Importing from the local FS is what most people seem to be doing.

Copying the source into HDFS is a reasonable alternative for now.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/tools/TransformTask.java
```diff
@@ -96,63 +96,32 @@ public TransformTask setNumWriters(int numWriters) {
...
   public PipelineResult run() throws IOException {
-    boolean runInParallel = true;
-    if (isLocal(from.getDataset()) || isLocal(to.getDataset())) {
-      runInParallel = false;
-    }
-
     PType<T> toPType = ptype(to);
     MapFn<T, T> validate = new CheckEntityClass<T>(to.getType());
 
-    if (runInParallel) {
```
rdblue: I think we need to explicitly set the mapreduce framework to local if either end is local. Here's what I ended up with:

```java
    if (isLocal(from.getDataset()) || isLocal(to.getDataset())) {
      getConf().set("mapreduce.framework.name", "local");
    }
```

kite-tools-parent/kite-tools/src/test/java/org/kitesdk/cli/commands/TestCSVImportCommand.java
```diff
@@ -196,15 +196,10 @@ public void testIncompatibleSchemaFieldType() throws Exception {
 
     // This will fail because NaN isn't a valid long and the field is required
     command.targets = Lists.newArrayList("target/incompatible.csv", datasetName);
-    TestHelpers.assertThrows("Should complain about schema compatibility",
```
rdblue: I'm fine merging this with this test simply ignored. We should add a follow-up to show the error message correctly.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/tools/TransformTask.java
```diff
@@ -96,63 +96,32 @@ public TransformTask setNumWriters(int numWriters) {
...
 
-      StageResult sr = Iterables.getFirst(result.getStageResults(), null);
-      if (sr != null && MAP_INPUT_RECORDS != null) {
-        this.count = sr.getCounterValue(MAP_INPUT_RECORDS);
-      }
-
-      return result;
-
-    } else {
-      Pipeline pipeline = MemPipeline.getInstance();
```
rdblue: Nit: It looks like when we remove the `MemPipeline` code, quite a few imports can be removed at the same time.

rdblue:
I've fixed some of the problems I ran into with this while testing, see #359.

rdblue:
These changes were merged as part of #359. Thanks Tom!

PR #304, opened 2015-02-09
CDK-885: Fix FS writer error handling.

rdblue:
This updates the FileSystemWriter's error handling. First,
RuntimeException is now caught throughout the class to ensure that
unknown exceptions in appenders (or FS calls) cannot put the writer in
an unknown state. Second, this updates the close method so that it
handles more error cases correctly:

* After a call to close, the writer state is always CLOSED
* Close does nothing when writer is NEW or CLOSED
* Close is called (on Appender) only if the writer is not in ERROR
* A data file is released if:
  1. Data has been written to the file
  2. Data has been committed with a flush() or sync()
  3. The writer is not in an ERROR state
* If a data file is not to be released, it is discarded
* Appender cleanup is called for OPEN or ERROR
* Any failure in close will prevent the next close task
* No close tasks will be retried b/c state is CLOSED

This also updates the TestFileSystemWriters tests and adds tests for
Parquet, which were previously missing.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetException.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetException.java
@@ -46,4 +46,11 @@ public class DatasetException extends RuntimeException {
     super(t);
   }
 
+  protected static String format(String message, Object... args) {
+    String[] argStrings = new String[args.length];
+    for (int i = 0; i < args.length; i += 1) {
+      argStrings[i] = String.valueOf(args[i]);
+    }
+    return String.format(String.valueOf(message), (Object[]) argStrings);
+  }
 }
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemWriter.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemWriter.java
@@ -21,7 +21,6 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableSet;
 import java.io.Closeable;
-import java.io.Flushable;
 import java.io.IOException;
 import java.util.Set;
 import java.util.UUID;
@@ -32,10 +31,13 @@ import org.apache.hadoop.mapreduce.RecordWriter;
 import org.apache.hadoop.mapreduce.TaskAttemptContext;
 import org.kitesdk.data.DatasetDescriptor;
 import org.kitesdk.data.DatasetIOException;
+import org.kitesdk.data.DatasetOperationException;
 import org.kitesdk.data.DatasetRecordException;
 import org.kitesdk.data.DatasetWriterException;
+import org.kitesdk.data.Flushable;
 import org.kitesdk.data.Format;
 import org.kitesdk.data.Formats;
+import org.kitesdk.data.Syncable;
 import org.kitesdk.data.ValidationException;
 import org.kitesdk.data.spi.AbstractDatasetWriter;
 import org.kitesdk.data.spi.DescriptorUtil;
@@ -59,26 +61,30 @@ class FileSystemWriter<E> extends AbstractDatasetWriter<E> {
     ));
   }
 
-  static interface FileAppender<E> extends Flushable, Closeable {
+  static interface FileAppender<E> extends java.io.Flushable, Closeable {
     public void open() throws IOException;
     public void append(E entity) throws IOException;
     public void sync() throws IOException;
     public void cleanup() throws IOException;
   }
 
-  protected final FileSystem fs;
   private final Path directory;
   private final DatasetDescriptor descriptor;
-  private FileAppender<E> appender;
   private Path tempPath;
   private Path finalPath;
-  private ReaderWriterState state;
   private int count = 0;
+
+  protected final FileSystem fs;
+  protected FileAppender<E> appender;
+  protected boolean flushed = false;
+
+  @VisibleForTesting
+  ReaderWriterState state;
 
   @VisibleForTesting
   final Configuration conf;
 
-  public FileSystemWriter(FileSystem fs, Path path, DatasetDescriptor descriptor) {
+  private FileSystemWriter(FileSystem fs, Path path, DatasetDescriptor descriptor) {
     Preconditions.checkNotNull(fs, "File system is not defined");
     Preconditions.checkNotNull(path, "Destination directory is not defined");
     Preconditions.checkNotNull(descriptor, "Descriptor is not defined");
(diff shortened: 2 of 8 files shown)


kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestAvroWriter.java
```diff
@@ -16,17 +16,80 @@
...
+    // 1. A failed record with an IOException or unknown RuntimeException
+    // 2. A failed flush or sync for IncrementableWriters
+    fsWriter.state = ReaderWriterState.ERROR;
+
+    fsWriter.close();
+
+    FileStatus[] stats = fs.listStatus(testDirectory, PathFilters.notHidden());
+    Assert.assertEquals("Should contain a visible data file", 1, stats.length);
+
+    // TODO: This currently fails because the file needs to be recovered
```
rdblue: @tomwhite, can we fix this easily?

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestFileSystemWriters.java
```diff
@@ -52,35 +63,74 @@ public void tearDown() throws IOException {
...
     fsWriter.close();
-    Assert.assertEquals("Should not contain any files", 0,
-        ImmutableList.copyOf(fs.listStatus(testDirectory)).size());
+
+    FileStatus[] stats = fs.listStatus(testDirectory, PathFilters.notHidden());
+    Assert.assertEquals("Should not contain any files", 0, stats.length);
   }
 
   @Test
-  public void testWrite() throws IOException {
```
rdblue: This test used AvroAppender directly, even when testing Parquet. I've updated it to test Parquet and Avro, as testBasicWrites

kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetException.java
```diff
@@ -46,4 +46,11 @@ public DatasetException(Throwable t) {
     super(t);
   }
 
+  protected static String format(String message, Object... args) {
+    String[] argStrings = new String[args.length];
+    for (int i = 0; i < args.length; i += 1) {
+      argStrings[i] = String.valueOf(args[i]);
+    }
+    return String.format(String.valueOf(message), (Object[]) argStrings);
```
tomwhite: What does this provide over `String.format(message, args)`? Message shouldn't be null. Also, if any of the args are null then they can be handled by the conversion specified in http://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html#syntax

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemWriter.java
```diff
@@ -148,64 +169,74 @@ public final void write(E entity) {
...
+  @Deprecated
   public void flush() {
-    Preconditions.checkState(state.equals(ReaderWriterState.OPEN),
-        "Attempt to flush a writer in state:%s", state);
-    try {
-      appender.flush();
-    } catch (IOException e) {
-      this.state = ReaderWriterState.ERROR;
-      throw new DatasetWriterException("Failed to flush appender " + appender);
-    }
```
tomwhite: Why has the implementation been removed here? Shouldn't it stay until after 0.18?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemWriter.java
```diff
@@ -148,64 +169,74 @@ public final void write(E entity) {
...
+  @Deprecated
   public void sync() {
-    Preconditions.checkState(state.equals(ReaderWriterState.OPEN),
-        "Attempt to sync a writer in state:%s", state);
-    try {
-      appender.sync();
-    } catch (IOException e) {
-      this.state = ReaderWriterState.ERROR;
-      throw new DatasetIOException("Failed to sync appender " + appender, e);
-    }
```
tomwhite: Ditto.

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestAvroWriter.java
tomwhite: Yes. This is failing because we're using the local filesystem, and hflush() isn't implemented for it (https://issues.apache.org/jira/browse/HADOOP-7844). I tried switching TestFileSystemWriters to extend MiniDFSTest and the commented out test passes.

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestAvroWriter.java
```diff
@@ -16,17 +16,80 @@
...
   }
+
+  @Override
+  public DatasetReader<Record> newReader(Path path, Schema schema) {
+    return new FileSystemDatasetReader<Record>(fs, path, schema, Record.class);
+  }
+
+  @Test
+  public void testIsFlushable() {
+    Assert.assertTrue(fsWriter instanceof FileSystemWriter.IncrementalWriter);
```
tomwhite: Test for the public interface (org.kitesdk.data.Flushable) rather than the private implementation class.

tomwhite:
I've added some comments, but overall this is a good improvement.

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestAvroWriter.java
rdblue: That doesn't work right now because it is always true (that was the original test). Both writers implement Flushable and Syncable, but only IncrementalWriter will in the future. I'll change this over when we remove Flushable and Syncable.

I've added a note to do this follow-up on [CDK-892](https://issues.cloudera.org/browse/CDK-892).

tomwhite: You're right. Fine to follow up.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetException.java
rdblue: First, I like being a little more careful with error messages so that any null values end up being "null" rather than causing a different exception and accidentally suppressing the one that matters. That's not too likely, though.

The other reason is to try to match the style of preconditions, which always use `%s`. It's annoying to need to remember whether the formatter for a method uses `{}`, `%s`, or allows `%d` etc. Plus that turns into a very bad thing when you get it wrong because it's a runtime bug and you usually lose key error information.

tomwhite: I think this can all be achieved by making the contents of this method `String.format(message, args)` since that will take care of formatting null (e.g. when you use `%s`).

Not a big deal either way though - this change shouldn't block the overall merge.

PR #310, opened 2015-02-10
CDK-372. Fix exception consistency and use DatasetIOException.

tomwhite:


kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/InputFormatReader.java
```diff
@@ -164,14 +164,14 @@ private boolean advance() {
...
-      throw new DatasetReaderException("Cannot advance reader", e);
+      throw new DatasetOperationException("Cannot advance reader", e);
     } catch (IOException e) {
       this.state = ReaderWriterState.ERROR;
       throw new DatasetIOException("Cannot advance reader", e);
     } catch (InterruptedException e) {
       // error: it is unclear whether the underlying reader is valid
       this.state = ReaderWriterState.ERROR;
-      throw new DatasetReaderException("Interrupted", e);
+      throw new DatasetOperationException("Interrupted", e);
```
rdblue: This should call `Thread.currentThread().interrupt()` rather than swallowing the interrupt.

http://www.ibm.com/developerworks/library/j-jtp05236/

rdblue:
One minor thing, but this looks good otherwise.

tomwhite:
I made the change to interrupt the current thread. Merged in https://github.com/kite-sdk/kite/commit/6d28bdfb8fc5989812b74c445575714afd9a00c6

PR #311, opened 2015-02-10
CDK-696: Store Avro schema and schema history in HDFS

rbrush:
Changes discussed in CDK-696: https://issues.cloudera.org/browse/CDK-696

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/SchemaManager.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/SchemaManager.java
@@ -1,0 +1,320 @@
+/**
+ * Copyright 2013 Cloudera Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kitesdk.data.spi.filesystem;
+
+import com.google.common.base.Charsets;
+import com.google.common.io.Closeables;
+import org.apache.avro.Schema;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FSDataOutputStream;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.kitesdk.data.DatasetException;
+import org.kitesdk.data.DatasetIOException;
+import org.kitesdk.data.IncompatibleSchemaException;
+import org.kitesdk.data.spi.SchemaValidationUtil;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.util.*;
+
+/**
+ * Manager for creating, updating, and using Avro schemas stored in a
+ * filesystem, typically HDFS.
+ */
+public class SchemaManager {
+
+  private final Path schemaDirectory;
+  private final FileSystem rootFileSystem;
+
+  /**
+   * Creates a new schema manager using the given root directory of a
+   * dataset for its base.
+   *
+   * @param conf the Hadoop configuration
+   * @param schemaDirectory directory in which the manager
+   *                        stores schemas.
+   *
+   * @return a schema manager instance.
+   */
+  public static SchemaManager create(Configuration conf, Path schemaDirectory) {
+
+    try {
+      FileSystem rootFileSystem = schemaDirectory.getFileSystem(conf);
+
+      rootFileSystem.mkdirs(schemaDirectory);
+
+      return new SchemaManager(schemaDirectory, rootFileSystem);
+    } catch (IOException e) {
+      throw new DatasetIOException("Unable to create schema manager directory: "
+              + schemaDirectory, e);
+    }
+
+  }
+
+  /**
+   * Loads a schema manager that stores data under the given dataset root
+   * directory it exists. Returns <code>null</code> if it does not.
+   *
+   * @param conf the Hadoop configuration
+   * @param schemaDirectory directory in which the manager stores schemas.
+   *
+   * @return a schema manager instance, or <code>null</code> if the given
+   * directory does not exist.
+   */
+  public static SchemaManager load(Configuration conf, Path schemaDirectory) {
+
+    try {
+
+      FileSystem rootFileSystem = schemaDirectory.getFileSystem(conf);
+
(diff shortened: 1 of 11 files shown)


rbrush:
The above build failure looks to be caused by failures downloading dependencies unrelated to this change. For instance, here is the cause of one of them:

```
[ERROR] Failed to execute goal on project kite-hbase-dependencies: Could not resolve dependencies for project org.kitesdk:kite-hbase-dependencies:pom:0.17.2-SNAPSHOT: Could not transfer artifact org.apache.hbase:hbase-client:jar:0.98.3-hadoop2 from/to central (http://repo.maven.apache.org/maven2): GET request of: org/apache/hbase/hbase-client/0.98.3-hadoop2/hbase-client-0.98.3-hadoop2.jar from central failed: Connection reset
```

rdblue:
Those are transient errors in Travis CI. I'll re-run any that are failing on transfers from maven central.

kite-data/kite-data-hive/src/main/java/org/kitesdk/data/spi/hive/HiveAbstractMetadataProvider.java
```diff
@@ -109,7 +112,25 @@ public DatasetDescriptor update(String namespace, String name, DatasetDescriptor
...
+
+      SchemaManager manager = SchemaManager.create(conf, managerPath);
+
+      DatasetDescriptor newDescriptor;
+
+      try {
+        URI schemaURI = manager.writeSchema(descriptor.getSchema());
+
+        newDescriptor = new DatasetDescriptor.Builder(descriptor)
+            .schemaUri(schemaURI).build();
```
rdblue: Does this successfully replace at schema literal with a schema URI?

kite-data/kite-data-hive/src/main/java/org/kitesdk/data/spi/hive/HiveExternalMetadataProvider.java
```diff
@@ -82,6 +83,25 @@ public DatasetDescriptor create(String namespace, String name, DatasetDescriptor
           .build();
     }
 
+    if (descriptor.getSchemaUrl() == null) {
+
+      // No schema URL is provided, so create one based on the dataset.
+      Path managerPath = new Path(new Path(newDescriptor.getLocation()),
+          SCHEMA_DIRECTORY);
```
rdblue: Shouldn't this always create a SchemaManager? If the URL is in the manager then that's a good thing (is that case detected?). Otherwise, the right thing to do is to copy the Schema to a managed location and rely on that one. That addresses the concern that the schema is changed without going through Kite (since Kite currently has just a pointer).

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/SchemaManager.java
```diff
@@ -0,0 +1,305 @@
...
+   * @return A URI pointing to the written schema
+   */
+  public URI writeSchema(Schema schema) {
+
+    Path previous = newestFile();
+
+    // Ensure all previous schemas are compatible with the new one.
+    Map<Integer, Schema> schemas = getSchemas();
+
+    for (Schema oldSchema: schemas.values()) {
```
rdblue: Should this also check if the schema is already present in the manager as the latest version? If so, then return?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/SchemaManager.java
```diff
@@ -0,0 +1,305 @@
...
+      if (!SchemaValidationUtil.canRead(oldSchema, schema)) {
```
rdblue: It would be good to have a comment explaining why all previous versions are checked. I think this is because we have to check the case where deleted columns are not compatible with the new schema, right?

rdblue:
Ryan, this is really great! I made a couple of notes, but I think the code is solid. We just need to work through the implications of changing the on-disk format.

I think the on-disk format needs to be forward-compatible: older Kite clients should be able to read at least one version up to allow people to upgrade and then roll back if they hit a regression or for some other reason need to. For this, that would mean maintaining the current `schema.avsc` in `.metadata` at the latest version in the manager. I'm fine with the schemas folder living along-side the schema file for HDFS, although the current `schema.avsc` should always be added to the manager as the latest or verified as the latest. If this is implemented, I don't see a need to update the version number until we stop maintaining the schema.avsc file.

For Hive, I think this is already compatible because it uses the existing URI property. I do want to make sure that Hive picks up the change and that the literal property is deleted unless it is used.

kite-data/kite-data-hive/src/main/java/org/kitesdk/data/spi/hive/HiveAbstractMetadataProvider.java
rbrush: Yes, the underlying table has a schema URI, which is validated in one of the unit tests that creates a schema too large to fit in a literal

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/SchemaManager.java
rbrush: Good idea. We can do that.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/SchemaManager.java
rbrush: Yes, that's the reason. I'll add a comment to that effect.

kite-data/kite-data-hive/src/main/java/org/kitesdk/data/spi/hive/HiveExternalMetadataProvider.java
rbrush: Ah, I didn't understand the desired behavior of the provided schema URL, and had assumed there was some case where maintaining an external reference was desired. But I agree that copying it the manager makes the system more robust and will make that change.

rbrush:
Thanks for the feedback, Ryan. Agreed on all counts. I'll restore the logic that maintains the current schema.avsc so we're forward-compatible, along with other suggestions made inline. I might not be able go get to it today but should have time tomorrow.

rbrush:
Just pushed a commit that I think covers the comments on this thread.  

It looks like master changed a bit recently so unless there are objections I might rebase these changes as well. (I don't think blowing up the commit history on this branch is going to break anyone.)

rbrush:
Rebased against master and ran all tests.

rbrush:
The build failures again appear to be transient exceptions pulling Maven dependencies.

rdblue:
Restarted failing travis test cases.

rbrush:
Failed tests are dependency retrieval failures again.

rbrush:
Rebased against master to ensure everything is still in good shape. (It is.)

tomwhite:
This looks good to merge to me. It would be great if @rdblue could take a final look too. I'd like to get this in to 1.0 which we'd like to do in the next couple of days.

rdblue:
+1

I'm excited to get this in for the next release!

tomwhite:
Merged in https://github.com/kite-sdk/kite/commit/ee9ff7d152020f215267d029e9dc41c8a66b7a9a

PR #313, opened 2015-02-11
CDK-766: Add CSV header CLI option.

rdblue:
This adds --header to csv-schema and csv-import. The option is a single
string that is passed to the CSV code in kite-data-core in
CSVProperties. The header can be stored as a new descriptor property,
kite.csv.header.

The CSVProperties header is always low-precedence. If useHeader is set
in properties, a header line is read from the file or data sample (for
inferSchema) and used instead. For CSVRecordParser, the properites
header is only used if the header passed is null.

The header line is always parsed using the other properties. There is no
special header CSV delimiter or other options.


tomwhite:
+1 I tried using the `csv-schema` command with a `--header` option and it worked as expected.

Merged in https://github.com/kite-sdk/kite/commit/3d42dbb41eeaf1df21f6173505e9baf8c5a972d7

PR #314, opened 2015-02-11
CDK-909. Morphlines Solr modules do not work on Java 6.

tomwhite:


whoschek:
I think you'd only need to exclude kite-morphlines-solr-core and kite-morphlines-solr-cell for java6, the other morphline modules are still ok with java6.

whoschek:
+1 Thanks!

PR #315, opened 2015-02-12
CDK-892. Remove DatasetWriter#flush and #sync

tomwhite:


kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemView.java
```diff
@@ -89,7 +89,7 @@ private FileSystemView(FileSystemView<E> view, Constraints c) {
   public DatasetWriter<E> newWriter() {
     AbstractDatasetWriter<E> writer;
     if (dataset.getDescriptor().isPartitioned()) {
-      writer = new PartitionedDatasetWriter<E>(this);
+      writer = PartitionedDatasetWriter.newWriter(this);
```
rdblue: Maybe `FileSystemWriter.newWriter` should handle this case as well?

rdblue:
+1

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/PartitionedDatasetWriter.java
```diff
@@ -259,4 +237,49 @@ public void onRemoval(
...
+    /*
+     * There's a potential for flushing entries that are created by other
+     * threads while looping through the writers. While normally just wasteful,
+     * on HDFS, this is particularly bad. We should probably do something about
+     * this, but it will be difficult as Cache (ideally) uses multiple
+     * partitions to prevent cached writer contention.
+     */
+      for (FileSystemWriter<E> writer : cachedWriters.asMap().values()) {
+        LOG.debug("Flushing partition writer:{}", writer);
+        if (writer instanceof Flushable) {
```
joey: Is it an error if one of our cached writers isn't Flushable?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/PartitionedDatasetWriter.java
```diff
@@ -259,4 +237,49 @@ public void onRemoval(
...
+    @Override
+    public void sync() {
+      Preconditions.checkState(state.equals(ReaderWriterState.OPEN),
+          "Attempt to sync a writer in state:%s", state);
+
+      LOG.debug("Syncing all cached writers for view:{}", view);
+
+      for (FileSystemWriter<E> writer : cachedWriters.asMap().values()) {
+        LOG.debug("Syncing partition writer:{}", writer);
+        if (writer instanceof Syncable) {
```
joey: Same question for syncable.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/PartitionedDatasetWriter.java
rdblue: Good point. Yes, I think it is an error. It would be better if the cache was parameterized with IncrementalWriter for the incremental case to avoid type checking, even if it is inserted automatically with a cast.

tomwhite: I've updated the code to avoid the instanceof test.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemView.java
tomwhite: Which case do you mean? FileSystemWriter.newWriter and PartitionedDatasetWriter.newWriter have the same logic now.

rdblue: Sorry, I wasn't clear. I mean that perhaps we should have a single utility method to return a writer for a view (partitioned or not) rather than having one for the partitioned case and one for the single-file case. Definitely not a blocker.

joey:
+1

PR #319, opened 2015-02-20
CDK-924: update the Hive metastore when partitions are removed.

rbrush:
The partition listener now listens to partition removal events and updates the Hive metastore appropriately.


rbrush:
Travis build failure caused by the usual failure getting dependencies.

PR #320, opened 2015-02-21
CDK-917: Add merge option to schema command.

rdblue:
This works with one or more schema args, which could be:
* A dataset or view URI
* A hive table name
* A resource URI
* A file path or HDFS path


tomwhite:
+1

rdblue:
Merged. Thanks for the review, Tom!

PR #321, opened 2015-02-22
CDK-902: Extend create command for existing data.

rdblue:
This adds a wrap command to create dataset metadata around existing data. Avro, Parquet, and JSON are supported (CSV has too many options to infer). Before committing this, the inspection methods should be moved to a utility library in core rather than in kite-tools.


kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/Schemas.java
```diff
@@ -0,0 +1,214 @@
...
+import org.apache.avro.generic.GenericDatumReader;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import parquet.avro.AvroSchemaConverter;
+import parquet.hadoop.ParquetFileReader;
+import parquet.hadoop.metadata.ParquetMetadata;
+
+public class Schemas {
```
rdblue: This class has more methods than we really need, with the intent of putting it in the API sometime soon. I think it makes more sense to use these methods with the descriptor builder rather than only exposing this functionality through the bulider's methods. Specifically, I'd rather see `builder.schema(Schemas.fromParquet(path))` rather than `builder.schemaFromParquetDataFile(path)`.

rdblue:
I think this is actually ready for review now. Thanks!

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/Schemas.java
```diff
@@ -0,0 +1,214 @@
...
+    return fromParquet(
+        FileSystem.getLocal(DefaultConfiguration.get()),
+        new Path(location.getPath()));
+  }
+
+  public static Schema fromParquet(FileSystem fs, Path location) throws IOException {
+    ParquetMetadata footer = ParquetFileReader.readFooter(fs.getConf(), location);
+
+    String schemaString = footer.getFileMetaData()
+        .getKeyValueMetaData().get("parquet.avro.schema");
```
joey: Parquet doesn't provide these property names as constants somewhere?

pom.xml
```diff
@@ -112,7 +112,7 @@
 
     <!-- Java versions -->
     <javaVersion>1.7</javaVersion>
-    <targetJavaVersion>1.6</targetJavaVersion>
+    <targetJavaVersion>1.7</targetJavaVersion>
```
joey: This should not be in the PR. Unless support was dropped for Hadoop 1 and CDH4, we can't rev the targetJavaVersion.

joey:
-1 for javaTargetVersion rev.

pom.xml
rdblue: Oops. I'll fix it.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/Schemas.java
rdblue: We will need to wait for 1.6 to reliably use Parquet's properties. Some versions used parquet.avro.schema and some used avro.schema. The current version, 1.5, only has a property for parquet.avro.schema because it doesn't have the compatibility for the older property. When the version with OLD_PARQUET_READ_SCHEMA is released we can update this.

rdblue:
Fixed the java version. Thanks for taking a look, Joey!

kite-data/kite-data-core/src/main/java/org/kitesdk/data/PartitionStrategy.java
```diff
@@ -160,6 +163,56 @@ PartitionStrategy getSubpartitionStrategy(int startIndex) {
         fieldPartitioners.size()));
   }
 
+  public void checkCompatibleUpdate(PartitionStrategy other, Schema schema) {
```
tomwhite: This shouldn't really be exposed as a public method. Put it on a class in the SPI?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/Schemas.java
tomwhite: I think it's confusing to have Schemas and SchemaUtil. Why not put these methods in SchemaUtil? Or name this SchemaInferer or something similar.

tomwhite:
This looks good to me, +1.

It would be good to have a test that ensures that trying to create an existing dataset (i.e. one with Kite metadata) fails, rather than inferring the schema and overwriting (I don't think it does this, but a test would be good).

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/Schemas.java
rdblue: If this is moved to the public API as a follow-up, then I think `Schemas` is the right name for it. Otherwise you're right and we should probably rename it or move it. What do you think about making this public?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/PartitionStrategy.java
rdblue: Will do.

rdblue:
Thanks for taking a look, Tom! I'll add the tests you suggest, good call. I think we should also validate that the partitions are actually added to the MetaStore, so I'll add a test for that before I merge.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/Schemas.java
tomwhite: No need to make this public - just leave it in the SPI for now.

rdblue:
Merged! Thanks for the review, Tom!

PR #324, opened 2015-02-22
CDK-674. DatasetDescriptor.checkPartitionStrategy should throw Validatio...

tomwhite:
...nException.

I changed methods in DatasetDescriptor, PartitionStrategy, and Compatibility to throw ValidationException rather than IllegalArgumentException or IllegalStateException when an invalid value is specified.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetDescriptor.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetDescriptor.java
@@ -905,7 +905,7 @@ public class DatasetDescriptor {
      * @since 0.9.0
      */
     public DatasetDescriptor build() {
-      Preconditions.checkState(schema != null,
+      ValidationException.check(schema != null,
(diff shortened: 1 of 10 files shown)


kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetDescriptor.java
```diff
@@ -982,7 +981,7 @@ private static void checkPartitionStrategy(
         }
 
         // check the entity is a record if there is a non-provided partitioner
-        Preconditions.checkState(schema.getType() == Schema.Type.RECORD,
+        ValidationException.check(schema.getType() == Schema.Type.RECORD,
```
rdblue: While this was already a problem, should we switch to passing objects rather than building the string here? This serializes the schema every time the function runs.

rdblue:
One minor comment, otherwise +1.

I also restarted the tests that had errors, but please check that they pass.

PR #327, opened 2015-02-24
CDK-927: Use requested type as reader schema in MapReduce jobs.

rbrush:
If a Dataset is loaded with a specific Avro type provided by the caller, that type is effectively the reader schema and should be honored by MapReduce jobs. This PR simply sets the reader schema appropriately in the input format and includes a test to ensure it is used.

### DIFF
--- a/kite-data/kite-data-crunch/pom.xml
+++ b/kite-data/kite-data-crunch/pom.xml
@@ -31,7 +31,19 @@
   </description>
 
   <build>
+    <resources>
+      <resource>
+        <directory>src/test/avro</directory>
+      </resource>
+      <resource>
+        <directory>src/main/resources</directory>
+      </resource>
+    </resources>
     <plugins>
+      <plugin>
+        <groupId>org.apache.avro</groupId>
+        <artifactId>avro-maven-plugin</artifactId>
+      </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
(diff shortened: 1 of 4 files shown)


rbrush:
Just realized we'll need to make a similar update for Parquet support, so this PR isn't complete. I'm not as familiar with the Parquet flow but will take a look.

rbrush:
Updated PR to handle this for Parquet inputs as well

tomwhite:
This looks good to me. It seems to have broken a unit test though.

tomwhite:
Actually the broken test may have been from https://issues.cloudera.org/browse/CDK-934

rbrush:
Yes, master is failing with this error as well, so I don't think it's this pull request that triggered that failure.

kite-data/kite-data-crunch/pom.xml
```diff
@@ -31,8 +31,20 @@
   </description>
 
   <build>
+    <resources>
+      <resource>
+        <directory>src/test/avro</directory>
```
rdblue: Why add src/test/avro to resources? Won't this include the schemas in the main jar?

rbrush: I might be guilty of cargo culting here, but I pulled this pattern from kite-data-core, which has the same stanza for its avro test resources. I verified the schema is not included in the resulting jar, so it looks like the jar step is smart enough to not include items from this folder.

There may be a better way to accomplish this, but it is at least consistent with other Kite sub-projects and produces the desired outcome.

rdblue: Okay, so kite-data-core does it so that the test schemas are included in the test-jar artifact. So it looks like this just includes it in the test jar correctly and not in the overall jar, so it is fine. Thanks!

rdblue:
+1 assuming tests pass.

PR #328, opened 2015-02-24
CDK-935: Fix log4j test for new Hive URL behavior.

rdblue:
Hive no longer uses the schema URL property if the schema is not in
HDFS. This updates the log4j test, removing the schema URL configuration
and adding the expected warning when not using a schema URL.


tomwhite:
You beat me to it - I was just writing the same fix.

+1

rdblue:
Merged

PR #329, opened 2015-02-24
CDK-937: Fix findbugs error.

rdblue:
The error is that datasets may be null and is always (not conditionally)
dereferenced. This is actually checked by a precondition so the code is
safe.


tomwhite:
+1

PR #330, opened 2015-02-27
CDK-944: Added a broken test to demonstrate this bug

joey:
I posted a fix to the bug exposed by the test case.


rdblue:
+1

Thanks, Joey!

PR #331, opened 2015-03-09
CDK-956. Allow nullable record field schemas in HBase.

tomwhite:


kite-data/kite-data-hbase/src/main/java/org/kitesdk/data/hbase/avro/AvroEntitySerDe.java
```diff
@@ -331,20 +332,32 @@ private void initKACRecordDatumMaps(String fieldName, Schema fieldSchema,
     }
   }
 
+  private static Schema unwrapNull(Schema schema) {
```
rdblue: Should this be named `unwrapNullable`?

Is it okay to assume that this is never called on a union other than of some type with null?

kite-data/kite-data-hbase/src/main/java/org/kitesdk/data/hbase/avro/AvroEntitySerDe.java
```diff
@@ -331,20 +332,32 @@ private void initKACRecordDatumMaps(String fieldName, Schema fieldSchema,
...
+      }
+    }
+    return schema;
+  }
+
   private DatumReader<Object> buildDatumReader(Schema schema,
       Schema writtenSchema) {
     if (specific) {
-      return new SpecificDatumReader<Object>(writtenSchema, schema);
+      return new SpecificDatumReader<Object>(unwrapNull(writtenSchema), unwrapNull(schema));
```
rdblue: I think this is okay because we are only going to call this reader if there is a value. But I'm not sure whether we are changing assumptions for other parts of the code. If fields were assumed to be non-null in the past, we should make sure the code checks whether the reader can handle nulls and sets the correct default value.

kite-data/kite-data-hbase/src/main/java/org/kitesdk/data/hbase/avro/AvroEntitySerDe.java
tomwhite: I fixed both of these in the latest.

kite-data/kite-data-hbase/src/main/java/org/kitesdk/data/hbase/avro/AvroEntitySerDe.java
tomwhite: Added a test for a null field.

joey:
I think we should add a test with a field of type `["string", "null"]` with a default `string` value to make sure that the right thing happens. I'm +1 after that test is added.

tomwhite:
Thanks Joey. Also, this could share the logic in https://issues.cloudera.org/browse/CDK-986 for unwrapping null schemas.

xguo27:
Are types ["string", "null"] and ["null", "string"] designed to be equivalent? I noticed that the generated java code is the same regardless of the order in the avro file.

If they are equivalent, are there any preference as for which format it is more conventional?

rdblue:
They are equivalent types, but when used as the type of a record field, the
field's default value has a restriction: its type must be the first one in
the union. So ["string", "null"] can have an empty default, "", while
["null", "string"] can only have a null default.

rb

On Wed, Mar 23, 2016 at 10:35 AM, Xiu Guo <notifications@github.com> wrote:

> Are types ["string", "null"] and ["null", "string"] designed to be
> equivalent? I noticed that the generated java code is the same regardless
> of the order in the avro file.
>
> If they are equivalent, are there any preference as for which format it is
> more conventional?
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/kite-sdk/kite/pull/331#issuecomment-200458036>
>



-- 
Ryan Blue

xguo27:
That explains. Thanks @rdblue !

PR #334, opened 2015-03-11
CDK-945: Ignore trailing slash in URIs.

rdblue:
This updates the URIPattern code to ignore trailing slashes in URIs.
URIs can still match a trailing / explicitly, and URI pattern globbing
still works as before.


tomwhite:
+1

PR #335, opened 2015-03-12
CDK-945: Improve error messages for URI problems.

rdblue:
* Add a suggestion to make sure default HDFS URI is configured
* Add known patterns for the matching URI scheme


tomwhite:
+1

PR #336, opened 2015-03-12
CDK-955: Fix duplicate URIs from Datasets.list.

rdblue:
The problem was that the FileSystem implementation was returning
multiple "default" namespaces. The implementation now uses a Set to
accumulate both namespaces and dataset names.

The bug only happened when a repository namespace path was used instead
of a repository. In that case, the single namespace appears to be a
valid repository with tables in the default namespace. Each table in the
default namespace would result in another "default" in the list of
namespaces. When Datasets.list listed the tables in each namespace, all
of the tables were present in each copy of default, which resulted in
n^2 results.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemMetadataProvider.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemMetadataProvider.java
@@ -295,8 +295,8 @@ public class FileSystemMetadataProvider extends AbstractMetadataProvider {
 
   @SuppressWarnings("deprecation")
   @Override
-  public List<String> namespaces() {
-    List<String> namespaces = Lists.newArrayList();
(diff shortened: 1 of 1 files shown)


kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemMetadataProvider.java
```diff
@@ -295,8 +295,8 @@ public boolean exists(String namespace, String name) {
 
   @SuppressWarnings("deprecation")
   @Override
-  public List<String> namespaces() {
-    List<String> namespaces = Lists.newArrayList();
+  public Set<String> namespaces() {
```
rdblue: So no one has to look this up, I confirmed that the interfaces require `Collection` so `Set` or `List` is fine.

joey:
+1, ship it!

rdblue:
Thanks, Joey!

PR #338, opened 2015-03-13
Corrected incorrect assertion message.

mkwhitacre:
Super trivial change to correct an assertion message that was copy/pasted incorrectly.


rdblue:
Thanks! Looks good to me, I'll merge in a bit.

PR #339, opened 2015-03-19
Add properties to ManagedSchema and ...

prazanna:
... copy over the Dataset properties to ManagedSchema during schema creation.
JIRA: https://issues.cloudera.org/browse/CDK-959

### DIFF
--- a/kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestFileSystemView.java
+++ b/kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestFileSystemView.java
@@ -266,7 +266,7 @@ public class TestFileSystemView extends TestRefinableViews {
     final Path nov11 = new Path("target/data/ns/test/year=2013/month=11/day=11");
     assertDirectoriesExist(fs, root, y2013, sep, sep12, oct, oct12, nov, nov11);
 
-    Assert.assertTrue("Delete should return false to indicate no changes",
+    Assert.assertTrue("Delete should return true to indicate data was deleted.",
         unbounded.deleteAll());
     assertDirectoriesDoNotExist(fs, y2013, sep12, sep, oct12, oct, nov11, nov);
     assertDirectoriesExist(fs, root);


kite-data/kite-data-hbase/src/main/avro/ManagedSchema.avsc
```diff
@@ -27,6 +27,9 @@
     { "name": "entitySchemas", "type": { "type": "map", "values": "string" },
       "mapping": { "type": "keyAsColumn", "value": "schema:" } },
 
+    { "name": "properties", "type": { "type": "map", "values": "string" },
+      "default" : null, "mapping": { "type": "column", "value": "meta:properties" } },
```
rdblue: The type isn't nullable, so I don't think you need the null default.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetDescriptor.java
```diff
@@ -935,6 +936,21 @@ public DatasetDescriptor build() {
         columnMapping = copiedMapping;
       }
 
+      // Parse for properties in the entity schema and merge them with
+      // properties defined. If a property is defined using this builder and
+      // is also in the schema, the value set in the builder will take precedence
+      if(EntityPropertiesParser.hasEmbeddedProperties(schema)) {
```
rdblue: This pattern was needed for the partition strategy and column mappings because we wanted to provide backward-compatibility for existing HBC schemas. Since properties are an addition to the SchemaManager, I don't think we need to worry about them being embedded in Schemas. I think that removes quite a bit of code!

kite-data/kite-data-hbase/src/main/java/org/kitesdk/data/hbase/HBaseMetadataProvider.java
```diff
@@ -168,6 +171,9 @@ public DatasetDescriptor update(String namespace, String name, DatasetDescriptor
     } else {
       LOG.info("Schema hasn't changed, not migrating: (" + name + ")");
     }
+
+    Map<String, String> properties = getProperties(descriptor);
```
rdblue: Do you think this should be a descriptor method? `descriptor.getProperties()` would make sense, along with a setter on the descriptor builder.

kite-data/kite-data-hbase/src/main/java/org/kitesdk/data/hbase/impl/SchemaManager.java
```diff
@@ -196,4 +196,55 @@ public void migrateSchema(String tableName, String entityName,
...
+
+  /**
+   * Sets the properties on the Managed Schema.
+   * All existing properties are removed.
+   *
+   * @param tableName  The table name of the managed schema
+   * @param entityName The entity name of the managed schema
+   * @param properties Map<String, String> properties
+   */
+  void setProperties(String tableName, String entityName,
```
rdblue: What about `getProperties(String table, String entity)` that returns `Map<String, String>`? Why not include it?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetDescriptor.java
prazanna: Hmm. When schema directory is migrated using the SchemaTool, either all the needed meta data is embedded inside the schema or I guess we would need a concept of a schema directory structure and define file names to read for these meta-data (paritions.json, schema.avro, entity.properties etc). The second approach is whole lot of work involved I guess. COTT team uses the migrateDirectory in SchemaTool to evolve schemas.

kite-data/kite-data-hbase/src/main/avro/ManagedSchema.avsc
prazanna: Yes. Removed.

kite-data/kite-data-hbase/src/main/java/org/kitesdk/data/hbase/HBaseMetadataProvider.java
prazanna: Yea that makes sense. But was not sure if that was intentionally avoided before with the listProperties() and getProperty() as the alternative. I will add a properties(Map<String, String> to the Descriptor builder and a getProperties() to the Descriptor

kite-data/kite-data-hbase/src/main/java/org/kitesdk/data/hbase/impl/SchemaManager.java
prazanna: Including it

rdblue:
Hey @prazanna, feel free to just force-push to your PR branch instead of opening a new PR each time. We don't mind force-pushes for that purpose. Thanks!

prazanna:
I actually did a force-push to my fork. But I am not sure about how this pull request would work on a forced push. Actually looks like it will readjust itself. I am going to try to reopen this.

PR #340, opened 2015-03-22
CDK-969: Support Thrift objects with avro-thrift.

rdblue:
This allows Thrift objects to be used via avro-thrift. Reads and writes
work when the dataset is backed by Avro. Parquet-avro currently assumes
that the classes implement IndexedRecord, which isn't the case when
using avro-thrift.


rdblue:
Tests are failing because of changes to the `schema(Class)` method, which now uses `AllowNull` instead of the regular `ReflectData`.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetDescriptor.java
```diff
@@ -486,7 +487,7 @@ public Builder schemaLiteral(String s) {
      * @since 0.2.0
      */
     public <T> Builder schema(Class<T> type) {
-      this.schema = ReflectData.get().getSchema(type);
+      this.schema = DataModelUtil.getTypeSchema(type);
```
joey: I agree this is the right thing to do, but does this break compatibility? The fact that it causes tests to fail I think means yes.

joey: A workaround might be to deprecate this method and add a new method in it's place.

rdblue: This breaks tests because it uses the `AllowNull` schema instead of assuming fields are generally non-null and annotated with `@Nullable` if necessary. This doesn't affect the behavior with any data already written, so it is a backward-compatible change. It is also a valid evolution because fields become more permissible. I think this would require CDK-986 (#353) but otherwise is pretty safe.

rdblue:
I should note that I haven't followed up with this very closely because it doesn't work with Parquet. At best, we could add the feature and not announce it, but I'm reluctant to add features that aren't working across both primary formats. That's one reason why I fixed reflect in Parquet, which I also need to follow up on.

kite-data/kite-data-core/src/test/java/org/kitesdk/data/event/thrift/Value.java
```diff
@@ -0,0 +1,427 @@
+/**
+ * Autogenerated by Thrift Compiler (0.7.0)
```
joey: I think we should check in the `.thrift` file that describes this type and add the thrift maven plugin to build this. Or maybe we can't because the maven plugin was added in thrift-0.9.2?

rdblue: The definition is checked in and I added a profile to regenerate this to the pom. The reason why I checked this in is to avoid making the build depend on having the thrift compiler installed, which would needlessly complicate our builds when all we really need is a single test file that won't change much.

joey: Ugh, I didn't know the maven plugin requires the executable installed. That's annoying.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/DatasetDescriptor.java
joey: But if it's breaking our unit tests doesn't that imply that it probably breaks users of the API?

joey:
> I should note that I haven't followed up with this very closely because it doesn't work with Parquet. At best, we could add the feature and not announce it, but I'm reluctant to add features that aren't working across both primary formats. That's one reason why I fixed reflect in Parquet, which I also need to follow up on.

I'm guessing that requires a future release of Parquet? If so, we probably have to defer this until 2.0 when we could update our minimum dependency version.

PR #341, opened 2015-03-22
CDK-938: Add kite-data-s3

rdblue:
This adds both s3n and s3a dataset URIs, s3n is working with a test.
Both URI patterns assume the bucket is the dataset repository.

Until a mock S3 service can be added, this expects real S3 credentials
and a bucket name passed as S3_ID, S3_KEY, and S3_BUCKET environment
variables. If those are not present, then the test will not run. To
disable tests for the hadoop1 and cdh4 profiles, the credentials are
explicitly set to empty.

### DIFF
--- a/kite-data/kite-data-s3/src/test/java/org/kitesdk/data/spi/s3/TestS3Dataset.java
+++ b/kite-data/kite-data-s3/src/test/java/org/kitesdk/data/spi/s3/TestS3Dataset.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright 2015 Cloudera Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


kite-data/kite-data-s3/src/test/java/org/kitesdk/data/spi/s3/TestS3Dataset.java
```diff
@@ -0,0 +1,147 @@
...
+import org.kitesdk.data.DatasetDescriptor;
+import org.kitesdk.data.DatasetReader;
+import org.kitesdk.data.DatasetWriter;
+import org.kitesdk.data.Datasets;
+import org.kitesdk.data.spi.DefaultConfiguration;
+
+public class TestS3Dataset {
+  private static final String ID = System.getProperty("aws.s3.id");
+  private static final String KEY = System.getProperty("aws.s3.key");
+  private static final String BUCKET = System.getProperty("aws.s3.bucket");
```
tomwhite: Put "test" in the property names to be clear that these are for testing only. Also, "key" is ambiguous, so spell out "aws.s3.access.key" and "aws.s3.secret.key" (to mirror s3a)?

tomwhite:
Looks good to me, pending one naming point.

kite-data/kite-data-s3/src/test/java/org/kitesdk/data/spi/s3/TestS3Dataset.java
rdblue: Will do. Thanks for taking a look!

tomwhite:
I've updated this in https://github.com/kite-sdk/kite/pull/356 with the renames, and also to allow a path in the S3 URI.

PR #342, opened 2015-03-24
Implement HBaseActionModifiable SPI to allow HBase level hookups into HB...

prazanna:
...ase Get/Put/Scan/Delete

### DIFF
--- a/kite-data/kite-data-hbase/src/main/java/org/kitesdk/data/hbase/DaoDataset.java
+++ b/kite-data/kite-data-hbase/src/main/java/org/kitesdk/data/hbase/DaoDataset.java
@@ -17,11 +17,20 @@ package org.kitesdk.data.hbase;
 
 import com.google.common.base.Preconditions;
 import java.net.URI;
+import java.util.List;
+
 import org.apache.avro.generic.IndexedRecord;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.mapreduce.InputFormat;
 import org.kitesdk.data.DatasetDescriptor;
 import org.kitesdk.data.Key;
+import org.kitesdk.data.hbase.impl.BaseDao;
+import org.kitesdk.data.hbase.impl.CompositeBaseDao;
+import org.kitesdk.data.hbase.impl.DeleteActionModifier;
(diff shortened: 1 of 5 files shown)


jackowaya:
This looks good to me. Other than copyrights and docs, the only thing that I would  consider is making unique modifiers available to get and delete as well. That's not super-critical though, which is why we haven't implemented it on the client template for the other methods. Probably more work than it's worth.

prazanna:
Corrected the Copyright year. Added Interface documentation (Actually missed to push a latest commit, the interface documentation was in that)

prazanna:
Thinking about the correct behavior of Action Modifiers. Ordering of action modifiers are important. I think an action modifier either replaces an equal one (equals()) in the list or gets added to the end. This provides some certainty while adding action modifiers without having to introspect the existing action modifiers. Unless I hear back, I will implement this and will update this PR.

rdblue:
@prazanna, what about using a name rather than equals to replace an existing modifier? That way you have more flexibility if you need to replace a modifier with one that isn't equal. This is minor.

prazanna:
@rdblue Hmm I preferred the equals() because it gives the flexibility. If I want the equals based on a name, I can implement it that way or I can just ignore equals() (or specify a name) if I dont care about duplicate modifiers. What do you think?

rdblue:
That sounds good to me. Like I said, it is a minor thing so whatever you want, as the person using it.

jackowaya:
I agree with @prazanna that equals is a little better here. 

I think we need to reword the documentation on the modifiers a little bit. It says that they run in the order that they are added, but it isn't clear from the doc that if you replace a modifier (by giving one that's equal to an existing one) your new modifier will run in the original one's slot.

The wording is a little tricky. Maybe instead of "in the order they are added" it should be "in the order they were originally added"? That's not great either.

prazanna:
@rdblue - If you are okay with the commit here, can you merge in the PR? Thanks.

rdblue:
+1

rdblue:
Merged in https://github.com/kite-sdk/kite/commit/005991d. Thanks @prazanna!

PR #343, opened 2015-03-26
CDK-970: Fix IncompatibleSchemaException.check.

rdblue:
Now throws IncompatibleSchemaException instead of ValidationException.


tomwhite:
Tests need updating, but otherwise looks good.

PR #344, opened 2015-03-26
CDK-971: Add method to find datasets from existing data files

rdblue:
This is based on CDK-902 and includes CDK-970 so tests pass. The relevant commit is https://github.com/kite-sdk/kite/commit/c819fc8c20ea5f4a69181bed3ac500e78db688cc


rdblue:
Fixed Tom's comments from https://github.com/rdblue/kite/commit/c819fc8c

This still depends on #343 and #321, so it isn't ready to go yet. I think it is ready for more reviews. This clarified the `findPotentialDatasets` behavior in docs: it purposely does not check external metadata sources, like the Hive MetaStore or `.metadata` folders. Instead, this shows what is actually in the FS.

This also sets a flag on the descriptor when a dataset has data files at different depths: "kite.filesystem.mixed-depth". Kite doesn't support mixed-depth datasets, but if the data appears to form a single dataset then this should report it correctly.

PR #345, opened 2015-03-31
CDK-978: Allow HBaseDatasetRepository to be created with an existing HTablePool

prazanna:
Allow HBaseDatasetRepository to be created with an existing HTablePool


rdblue:
`HTablePool` is removed in newer versions, should we stop using it instead of patching the behavior? I guess this is now the recommended method: https://hbase.apache.org/book.html#client.connections

prazanna:
Yea, thats probably best. Let me see if I can find time to do that. Closing this PR.

PR #346, opened 2015-04-01
CDK-973: Added support for setting a reader schema in MR

joey:
* Added support for setting a reader schema for crunch sources
* Updated the copytask to set the reader schema to the schema of the
  destination dataset.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/View.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/View.java
@@ -17,6 +17,8 @@ package org.kitesdk.data;
 
 import java.net.URI;
 import javax.annotation.concurrent.Immutable;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericRecord;
 
 /**
  * A {@code View} is a subset of a {@link Dataset}.
@@ -109,6 +111,15 @@ public interface View<E> {
   public Class<E> getType();
 
   /**
+   * Get the schema of entities contained in this {@code View}.
+   *
+   * @return the schema of entities contained in this view
+   *
+   * @since 1.1.0
+   */
+  public Schema getSchema();
+
+  /**
    * Check whether this {@link View} contains any records.
    *
    * Implementations should return once a single record in this view is found.
@@ -128,4 +139,18 @@ public interface View<E> {
    * @since 0.16.0
    */
   public URI getUri();
+
+  /**
+   * Creates a copy of this {@code View} that projects entities to the given
+   * {@link Schema}.
+   * <p>
+   * This method always returns a {@code View} with type {@link GenericRecord}.
+   *
+   * @param <T> the type of {@code GenericRecord} to use
+   * @param schema an Avro schema to project entities to
+   * @return a copy of this view that projects entities to the given schema
+   * @throws IncompatibleSchemaException the given {@code schema} is incompatible
+   * with the underlying dataset.
+   */
+  <T extends GenericRecord> View<T> asSchema(Schema schema);
 }
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/AbstractDataset.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/AbstractDataset.java
@@ -23,6 +23,8 @@ import org.kitesdk.data.DatasetWriter;
 import org.kitesdk.data.RefinableView;
 import javax.annotation.concurrent.Immutable;
 import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericRecord;
+import org.kitesdk.data.View;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -39,9 +41,11 @@ public abstract class AbstractDataset<E> implements Dataset<E>, RefinableView<E>
 
   protected abstract RefinableView<E> asRefinableView();
   protected final Class<E> type;
+  protected final Schema schema;
 
   public AbstractDataset(Class<E> type, Schema schema) {
     this.type = DataModelUtil.resolveType(type, schema);
+    this.schema = DataModelUtil.getReaderSchema(this.type, schema);
   }
 
   @Override
@@ -76,6 +80,11 @@ public abstract class AbstractDataset<E> implements Dataset<E>, RefinableView<E>
   }
 
   @Override
+  public Schema getSchema() {
+    return schema;
+  }
+
+  @Override
   public RefinableView<E> with(String name, Object... values) {
     return asRefinableView().with(name, values);
   }
@@ -98,6 +107,11 @@ public abstract class AbstractDataset<E> implements Dataset<E>, RefinableView<E>
   @Override
   public RefinableView<E> toBefore(String name, Comparable value) {
     return asRefinableView().toBefore(name, value);
+  }
+
+  @Override
+  public <T extends GenericRecord> View<T> asSchema(Schema schema) {
+    return asRefinableView().asSchema(schema);
   }
 
   @Override
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/AbstractRefinableView.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/AbstractRefinableView.java
@@ -21,6 +21,7 @@ import com.google.common.base.Predicate;
 import java.net.URI;
 import java.util.Map;
 import javax.annotation.concurrent.Immutable;
+import org.apache.avro.Schema;
 import org.kitesdk.data.Dataset;
 import org.kitesdk.data.DatasetDescriptor;
 import org.kitesdk.data.DatasetReader;
@@ -74,6 +75,17 @@ public abstract class AbstractRefinableView<E> implements RefinableView<E> {
     this.entityTest = constraints.toEntityPredicate(accessor);
   }
 
+  protected AbstractRefinableView(AbstractRefinableView<?> view, Schema schema) {
+    this.dataset = (Dataset<E>) view.dataset;
+    this.comparator = view.comparator;
+    this.constraints = view.constraints;
+    // thread-safe, so okay to reuse when views share a partition strategy
+    this.keys = view.keys;
+    // Resolve our type according to the given schema
+    this.accessor = DataModelUtil.accessor(schema);
+    this.entityTest = constraints.toEntityPredicate(accessor);
+  }
+
   protected AbstractRefinableView(AbstractRefinableView<E> view, Constraints constraints) {
     this.dataset = view.dataset;
     this.comparator = view.comparator;
@@ -106,6 +118,11 @@ public abstract class AbstractRefinableView<E> implements RefinableView<E> {
   @Override
   public Class<E> getType() {
     return accessor.getType();
+  }
+
+  @Override
+  public Schema getSchema() {
+    return accessor.getEntitySchema();
   }
 
   public EntityAccessor<E> getAccessor() {
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DataModelUtil.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DataModelUtil.java
@@ -207,4 +207,38 @@ public class DataModelUtil {
     return new EntityAccessor<E>(type, schema);
   }
 
+  public static <E> E deepCopy(E entity, Schema dstSchema) {
+    GenericData dataModel = getDataModelForType(entity.getClass());
+    Schema srcSchema = dataModel.induce(entity);
+    if (dataModel instanceof ReflectData || dataModel instanceof SpecificData) {
+      return dataModel.deepCopy(dstSchema, entity);
+    } else {
+      if (srcSchema.equals(dstSchema)) {
+        return dataModel.deepCopy(dstSchema, entity);
+      }
+      GenericRecord srcRecord = (GenericRecord) entity;
+      return (E) deepCopy(srcRecord, srcSchema, dstSchema, dataModel);
+    }
+  }
+
+  private static GenericRecord deepCopy(GenericRecord srcRecord,
+    Schema srcSchema, Schema dstSchema, GenericData dataModel) {
+    GenericRecord dstRecord = (GenericRecord) dataModel.newRecord(null, dstSchema);
+    for (Schema.Field srcField : srcSchema.getFields()) {
+      Schema.Field dstField = dstSchema.getField(srcField.name());
+      if (dstField != null) {
+        if (srcField.schema().getType() == Schema.Type.RECORD) {
+          dstRecord.put(dstField.pos(),
+            deepCopy((GenericRecord)srcRecord.get(srcField.pos()),
+              srcField.schema(), dstField.schema(), dataModel));
+        } else {
+          dstRecord.put(dstField.pos(), dataModel.deepCopy(dstField.schema(),
+            srcRecord.get(srcField.pos())));
+        }
+      }
+    }
+
+    return dstRecord;
+  }
+
 }
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DataModelUtil.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DataModelUtil.java
@@ -101,6 +101,25 @@ public class DataModelUtil {
   }
 
   /**
+   * Returns true if the type implements GenericRecord but not SpecificRecord
+   *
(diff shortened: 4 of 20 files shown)


kite-data/kite-data-mapreduce/src/main/java/org/kitesdk/data/mapreduce/DatasetKeyOutputFormat.java
```diff
@@ -324,7 +324,7 @@ public void write(E key, Void v) {
     }
 
     private <E> E copy(E key) {
-      return dataModel.deepCopy(schema, key);
+      return DataModelUtil.deepCopy(key, schema);
```
joey: I'm not convinced we want this here, but I wanted a second opinion.

kite-data/kite-data-mapreduce/src/test/java/org/kitesdk/data/mapreduce/TestDatasetRecordWriter.java
```diff
@@ -0,0 +1,106 @@
...
+      record.put("name", "name"+count);
+
+      recordWriter.write(record, null);
+    }
+
+    recordWriter.close(null);
+  }
+  
+  @Test
+  public void testRecordWriterWithEvolution() {
```
joey: I'm not sure this is a valid test case. It seems fair that you should be producing entities with a schema that matches the destination when using `DatasetKeyOutputFormat`.

kite-tools-parent/kite-tools/src/test/java/org/kitesdk/cli/commands/TestCopyCommandCluster.java
```diff
@@ -142,16 +145,21 @@ public void testCopyWithoutCompaction() throws Exception {
...
   }
 
   @Test
-  @SuppressWarnings("unchecked")
   public void testCopyWithNumWriters() throws Exception {
+    testCopyWithNumWriters(3);
+  }
+
+  @SuppressWarnings("unchecked")
+  public void testCopyWithNumWriters(int expectedFiles) throws Exception {
```
joey: I'm not thrilled with these test changes, but they drastically cut down on code duplication for the new tests that I added.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DataModelUtil.java
```diff
@@ -207,4 +207,38 @@ public static Schema makeNullableSchema(Schema schema) {
     return new EntityAccessor<E>(type, schema);
   }
 
+  public static <E> E deepCopy(E entity, Schema dstSchema) {
```
rdblue: These may be good to have later, but if they aren't needed for this commit let's remove them. We can add them as a separate feature and add appropriate tests. Also, if what we want is a schema update then you might want to do that instead of a deep copy. There's no need to deep copy when the schema for a sub-record hasn't changed.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemViewKeyInputFormat.java
```diff
@@ -76,24 +77,27 @@ public FileSystemViewKeyInputFormat(FileSystemDataset<E> dataset,
     Format format = dataset.getDescriptor().getFormat();
 
     boolean isSpecific = SpecificRecord.class.isAssignableFrom(dataset.getType());
+    String readerSchema = conf.get("kite.readerSchema");
+    if (isSpecific && readerSchema == null) {
```
rdblue: I think that this should require that the schema is null if `isSpecific` is true. You should be able to project columns by using a particular class (specific, reflect, thrift, etc.) or by setting a schema, but not both. We should not return partial objects: when I get an object each field should be what was stored in the dataset. This is especially important because Avro uses unsafe APIs to construct reflected objects so there are fewer guarantees.

kite-data/kite-data-crunch/src/main/java/org/kitesdk/data/crunch/CrunchDatasets.java
```diff
@@ -96,6 +96,52 @@
...
+   * Expose the given {@link View} as a Crunch {@link ReadableSource}.
+   *
+   * @param view the view to read from
+   * @param readerSchema the schema used to read from the view
+   * @param <E>     the type of entity produced by the source
+   * @return a {@link ReadableSource} for the view
+   *
+   * @since 1.1.0
+   */
+  public static <E> ReadableSource<E> asSource(View<E> view, Schema readerSchema) {
```
rdblue: I don't think we should pass both read schema and view. That allows setting the expected class and a projection, which I don't think should happen. Instead, I think the right thing to do is to extend views to handle column projection. They are half-way there by carrying the expected type already and the intent at the time (at least mine) was to finish the API by adding more variations for column projection.

Here are some methods we could add:
* `View<?>#as(Schema) => View<GenericRecord>`
* `View<?>#as(Class<T>) => View<T>`
* `View<?>#select(String...) => View<GenericRecord>`
* `View<?>#select(List<String>) => View<GenericRecord>`

kite-data/kite-data-mapreduce/src/main/java/org/kitesdk/data/mapreduce/DatasetKeyInputFormat.java
```diff
@@ -137,6 +139,12 @@ public ConfigBuilder readFrom(String uri) {
       conf.setClass(KITE_TYPE, type, type);
       return this;
     }
+
+    public ConfigBuilder withSchema(Schema readerSchema) {
```
rdblue: I'd keep this, but add validation to make sure the class isn't set by a view. Also, the methods that take views should call it where appropriate. And possibly add `select(String...)`?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DataModelUtil.java
joey: Yeah, I can remove it.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemViewKeyInputFormat.java
joey: Yeah, we should only let you set a readerSchema if you're using GenericRecords.

kite-data/kite-data-crunch/src/main/java/org/kitesdk/data/crunch/CrunchDatasets.java
rdblue: Here's a little more detail: https://gist.github.com/rdblue/4f36915e2bc9a0261c65

joey: Do you think it's worth adding these APIs now just for this issue, or would it make sense to give the CopyTask a private API to solve the current problem and add more general support for this in a follow-up?

rdblue: I don't think we need anything but `View#asSchema` but I think we should do that instead of a band-aid solution.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemViewKeyInputFormat.java
joey: Should we be setting the reader schema if your data model is reflect or only for specific based on the type and for generic based on the reader schema?

Also, should we always be setting the reader schema? If the user doesn't set it, either by creating a view or by calling `withSchema` on the `DatasetKeyInputFormat`, we should probably set it to the schema of the dataset so that all records are read using the same schema, right?

rdblue: I think we currently default the reader schema to the writer schema for specific and reflect (see [`DataModelUtil`](https://github.com/kite-sdk/kite/blob/master/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DataModelUtil.java#L167)). That logic is correct for specific and generic, but reflect will need some special handling because there are two ways to interpret the records.

When reading, we should use the dataset's schema, but make sure that it is compatible with the reflected schema using AllowNull because that is what the objects can hold. When writing, we need to use the dataset's schema but validate that it can read the reflected schema without AllowNull or else we will have a problem.

joey: I would think with MapReduce, we'd want to use the Dataset schema as the reader schema. That way when we're reading from multiple files that were written with different schemas, everything is read with the same schema.

rdblue: To ensure the reader schema is consistent across all files, we should determine what it is before running the job. But it always needs to be based on the record that will hold the contents. For specific, it should be the specific record's schema and for reflect (reading) it should be the AllowNull schema of the requested type. For generic, it should default to the dataset schema, but can be overridden with a projected schema or a less restrictive schema.

joey: That was my thought as well. The current implementation doesn't do that. It only handles SpecificRecords and GenericRecords. I think I need to add support for setting the reader schema for ReflectRecords using DataModelUtil.

I can do that in this PR or in a follow-up JIRA.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/View.java
```diff
@@ -109,6 +111,15 @@
   public Class<E> getType();
 
   /**
+   * Get the schema of entities contained in this {@code View}.
+   *
+   * @return the schema of entities contained in this view
+   *
+   * @since 1.1.0
+   */
+  public Schema getSchema();
```
rdblue: I think this should be named `getReadSchema` instead. The problem is that a dataset is also a view, so we need a way to distinguish between the dataset's schema and the schema you get when using it as a view. Otherwise it appears that I can change the dataset's schema using `asSchema`.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/AbstractDataset.java
```diff
@@ -76,6 +80,11 @@ public boolean includes(E entity) {
   }
 
   @Override
+  public Schema getSchema() {
+    return schema;
```
rdblue: Should this use asRefinableView().getSchema()?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DataModelUtil.java
```diff
@@ -101,6 +101,25 @@ public static Schema makeNullableSchema(Schema schema) {
...
+   * @param <E> The entity type
+   * @param type The Java class of the entity type
+   * @return true if the type implements GenericRecord but not SpecificRecord
+   */
+  public static <E> boolean isGeneric(Class<E> type) {
+    // Need to check if SpecificRecord first because specific records also
+    // implement GenericRecord
+    if (SpecificRecord.class.isAssignableFrom(type)) {
+      return false;
+    } else if (IndexedRecord.class.isAssignableFrom(type)) {
```
rdblue: I think this should be GenericRecord instead of IndexedRecord. Indexed doesn't necessarily signal generic. You can also use this to simplify the logic in `createRecord`.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/View.java
rdblue: Okay, thinking about this more... I don't think my suggestion is valid.

The schema you set on the view is the view's schema and isn't necessarily the a read schema. The dataset's schema must always be able to read that schema, but it can be more strict that the dataset's schema. It can also be more permissive, in the case where you're projecting a subset of the columns you don't care about the ones that aren't being read.

So then we can't validate the schema from `asSchema` when it is used to create a view, only when we use that view to get a reader or a writer. (Which we should do, by the way.) Because the schema could be either a read or write schema, we can't name this method for either one. But we still need to distinguish it from the dataset's schema. `getViewSchema` or `getEffectiveSchema`?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemViewKeyInputFormat.java
```diff
@@ -76,24 +79,32 @@ public FileSystemViewKeyInputFormat(FileSystemDataset<E> dataset,
     Format format = dataset.getDescriptor().getFormat();
 
     boolean isSpecific = SpecificRecord.class.isAssignableFrom(dataset.getType());
+    String readerSchema = conf.get(KITE_READER_SCHEMA);
+
+    Preconditions.checkState(!isSpecific || readerSchema == null,
+      "Illegal configuration: {} must not be set if dataset uses a specific type {}",
+      KITE_READER_SCHEMA, dataset.getType());
```
rdblue: If the reader schema property is private and we don't allow the ConfigBuilder to set both, then I'm fine with using a warning here instead of failing the job, with the specific object taking precedence.

kite-data/kite-data-mapreduce/src/main/java/org/kitesdk/data/mapreduce/DatasetKeyInputFormat.java
```diff
@@ -134,9 +145,22 @@ public ConfigBuilder readFrom(String uri) {
...
+      String readerSchema = conf.get(KITE_READER_SCHEMA);
+      Preconditions.checkArgument(DataModelUtil.isGeneric(type) || readerSchema == null,
+        "Can't configure a type when a reader schema is already set: {}", readerSchema);
       conf.setClass(KITE_TYPE, type, type);
       return this;
     }
+
+    public ConfigBuilder withSchema(Schema readerSchema) {
+      Class<?> type = conf.getClass(KITE_TYPE, null);
+      Preconditions.checkArgument(type == null,
```
rdblue: It looks like `withType` allows setting a generic type. In that case, it would make sense to call both withType and withSchema.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/AbstractRefinableView.java
```diff
@@ -74,6 +75,17 @@ protected StorageKey initialValue() {
     this.entityTest = constraints.toEntityPredicate(accessor);
   }
 
+  protected AbstractRefinableView(AbstractRefinableView<?> view, Schema schema) {
+    this.dataset = (Dataset<E>) view.dataset;
```
rdblue: I don't think this works. If I had a dataset that was instantiated with a specific type, then called asSchema to change the schema, I don't end up with a usable dataset because it still has the same type class. I think this will have to create a copy of the dataset with GenericData.Record as its class.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/View.java
rdblue: Another option that potentially solves both this and the dataset cast issue below (need two copies of a dataset with different types) is to make the instance class type available only on views. Then the dataset simply has a schema that is the same as the descriptor's schema. You can then get a view with a different type or a view with a different schema, but the underlying dataset is always `Dataset<GenericRecord>`.

The `Datasets` class would be simpler, replacing all methods that take a type with something like this:

```java
View<Specific> data = Datasets.load(uri).asType(Specific.class);
```

kite-data/kite-data-mapreduce/src/test/java/org/kitesdk/data/mapreduce/TestDatasetRecordWriter.java
```diff
@@ -0,0 +1,72 @@
...
+    dataset = repo.create("ns", "out",
+        new DatasetDescriptor.Builder()
+            .property("kite.allow.csv", "true")
+            .schema(STATS_SCHEMA)
+            .format(format)
+            .build(), GenericData.Record.class);
+  }
+
+  @Test
+  public void testBasicRecordWriter() {
```
rdblue: I think we need a test for trying to write with a schema that can't be read. The call to `newWriter` should throw an `IncompatibleSchemaException`.

joey:
Closing in favor of @rdblue's PR

PR #349, opened 2015-04-13
Speed up schema migration when loading all schemas from a directory

prazanna:
The following changes are made
1. kicking off HbaseAdmin.createTableAsync for a bunch of tables and waiting for HBaseAdmin.isTableAvailable for all the tables is much quicker than the sync version HBaseAdmin.createTable sequentially for all the tables. The SchemaTool waits for a maximum of 10 minutes for all the tables to become available which seems like a very legible buffer.
2. HBaseAdmin.disableTable and HBaseAdmin.enableTable are very costly (~25% time spent over the schema migration). Instead of creating tables as and when we know about a entity schema, construct the HTableDescriptor for all the entity schemas for a specific table, this was we dont have to add column families at a later point of time which requires the disable and enable

### DIFF
--- a/kite-data/kite-data-hbase/src/main/java/org/kitesdk/data/hbase/avro/example/UserProfileExample.java
+++ b/kite-data/kite-data-hbase/src/main/java/org/kitesdk/data/hbase/avro/example/UserProfileExample.java
@@ -68,7 +68,7 @@ public class UserProfileExample {
    * The constructor will start by registering the schemas with the meta store
    * table in HBase, and create the required tables to run.
    */
-  public UserProfileExample() {
+  public UserProfileExample() throws InterruptedException {
     Configuration conf = HBaseConfiguration.create();
     HTablePool pool = new HTablePool(conf, 10);
     SchemaManager schemaManager = new DefaultSchemaManager(pool);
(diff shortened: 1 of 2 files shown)


rdblue:
Looks like this breaks the CDH4 and Hadoop 1 profiles. That's not necessarily a blocker because we need to discuss when to drop support for them. If that's the route you'd like to go, could you bring it up on the list? Otherwise, we should try to make this work for all profiles.

prazanna:
I trust I have fixed the hadoop-1 and cdh4 profiles now. But it looks like Travis has hit some kind of transient error building the hadoop-1 profile.

prazanna:
testValidMiniCluster(org.kitesdk.minicluster.TestMiniCluster)  Time elapsed: 3.805 sec  <<< ERROR!
java.net.BindException: Port in use: localhost:50070.

Can someone with write access re-kick the travis build for this PR? Thanks.

rdblue:
Kicked, thanks @prazanna!

rdblue:
Just merged this as https://github.com/kite-sdk/kite/commit/acd0459095fb095971183394067f3208cfebac60

I forgot, but in the future could you make sure the commit message starts with a reference to a JIRA issue? Thanks!

PR #350, opened 2015-04-13
CDK-982: Modified DatasetDescriptor updates to be atomic using FS renami...

mkwhitacre:
...ng.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemMetadataProvider.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemMetadataProvider.java
@@ -516,16 +516,17 @@ public class FileSystemMetadataProvider extends AbstractMetadataProvider {
       properties.setProperty(property, descriptor.getProperty(property));
     }
 
+    final Path tempDescriptorPath = new Path(metadataLocation, DESCRIPTOR_FILE_NAME+"_tmp");
     final Path descriptorPath = new Path(metadataLocation, DESCRIPTOR_FILE_NAME);
     threw = true;
     try {
-      outputStream = fs.create(descriptorPath, true /* overwrite */ );
+      outputStream = fs.create(tempDescriptorPath, true /* overwrite */ );
       properties.store(outputStream, "Dataset descriptor for " + name);
       outputStream.flush();
       threw = false;
     } catch (IOException e) {
       throw new DatasetIOException(
-          "Unable to save descriptor file: " + descriptorPath +
+          "Unable to save descriptor file: " + tempDescriptorPath +
           " for dataset: " + name, e);
     } finally {
       try {
@@ -534,6 +535,16 @@ public class FileSystemMetadataProvider extends AbstractMetadataProvider {
         throw new DatasetIOException("Cannot close", e);
       }
     }
+
+    //rename the file into its final location so that changes appear atomic
+      try {
+          fs.rename(tempDescriptorPath, descriptorPath);
+      } catch (IOException e) {
+          throw new DatasetIOException(
+                  "Unable to finalize descriptor file update to: " + descriptorPath +
+                          " for dataset: " + name, e);
+      }
+
   }
 
   /**


kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemMetadataProvider.java
```diff
@@ -534,6 +535,16 @@ static void writeDescriptor(
         throw new DatasetIOException("Cannot close", e);
       }
     }
+
+    //rename the file into its final location so that changes appear atomic
+      try {
+          fs.rename(tempDescriptorPath, descriptorPath);
```
ben-roling: Filesystem.rename() is a somewhat awkward API that has a boolean return value.  It doesn't work the way you might expect.  Interestingly the JavaDoc doesn't say anything about when it might return false but I don't think it will do the overwrite.  I believe what you need to do an atomic rename with overwrite is FileContext.rename(src, dest, Options.Rename.OVERWRITE).

mkwhitacre:
Thanks @ben-roling for the feedback.  Also looking through the SchemaManager class seems like it would be good to make a similar change there to make sure partially written files don't affect running code.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemMetadataProvider.java
```diff
@@ -516,16 +516,17 @@ static void writeDescriptor(
       properties.setProperty(property, descriptor.getProperty(property));
     }
 
+    final Path tempDescriptorPath = new Path(metadataLocation, DESCRIPTOR_FILE_NAME+"_tmp");
```
rdblue: The convention used elsewhere is to make this a dot file so it is hidden.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemMetadataProvider.java
```diff
@@ -534,6 +535,16 @@ static void writeDescriptor(
         throw new DatasetIOException("Cannot close", e);
       }
     }
+
+    //rename the file into its final location so that changes appear atomic
+      try {
```
rdblue: Nit: the formatting in this section is off. Formatting should use 2 spaces for an indent. (I know this is tedious, but it really helps with maintenance later.)

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemMetadataProvider.java
bbrownz: @ben-roling and I were chatting about this this morning. FileContext would do the rename you're expecting but wouldn't pass on the hadoop-1 profile. We're seeing that issue over on another PR now https://github.com/kite-sdk/kite/pull/348#discussion_r28332266

rdblue:
This looks like a good addition to me, but I'm not sure what guarantees we need to make for Schemas. Currently, the last schema in the schemas folder is considered current, and we keep a copy of it in `.metadata/schema.avsc`. Should a reference to the current schema be part of the descriptor so that a schema update is atomic as well?

Otherwise, I think we might be open to cases where a new schema is in the repository that isn't compatible with the partition strategy in the file. Maybe this isn't something we should worry about though.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemMetadataProvider.java
mkwhitacre: It's less than ideal but could use the deprecated rename method that takes an Option right?[1]

[1] - https://hadoop.apache.org/docs/current/api/org/apache/hadoop/fs/FileSystem.html#rename(org.apache.hadoop.fs.Path, org.apache.hadoop.fs.Path, org.apache.hadoop.fs.Options.Rename...)

ben-roling: Yeah, the only option I could see would potentially be going directly to the DFSClient API to call DFSClient.rename(src, dst, options).  I'm not sure what the implications of doing that would be though.  Certainly doesn't seem ideal.

mkwhitacre:
```
This looks like a good addition to me, but I'm not sure what guarantees we need to make for Schemas. Currently, the last schema in the schemas folder is considered current, and we keep a copy of it in .metadata/schema.avsc
```

So the code actually does the write twice, first in the .metadata/schema.avsc and then again in the schemas/ folder.  So this actually raises the problem where the .metadata/schema.avsc could be updated but the historical schema is not written so you lose that for schema compatibility checking.  

While retconning the compatibility with hadoop-1 how far back in Kite versions are you compatible?  The change I was thinking was excluding/dropping the .metadata/schema.avsc completely.  Then when writing always write out the file to be .metadata/schemas/.temp#.avsc and then rename once completed.  Then the load only looks for the newest schema.  If an update to the schema succeeds but descriptor fails since you know it is compatible it shouldn't affect anything.  It would just be a schema for data that could exist.

rdblue:
> The change I was thinking was excluding/dropping the .metadata/schema.avsc completely

I think we need to keep this for compatibility, but it wouldn't carry atomicity guarantees. In other words, lets ignore it in terms of this discussion. I think we can just be concerned about the schemas directory and the previous metadata conventions just continue to be not atomic.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemMetadataProvider.java
mkwhitacre: Another wrinkle in the FileContext route is that it even if hadoop-1 is dropped it doesn't look like webhdfs is supported..

When you switch to using it you get an an error building the project with a message like:

```
org.apache.hadoop.fs.UnsupportedFileSystemException: No AbstractFileSystem for scheme: webhdfs
	at org.apache.hadoop.fs.AbstractFileSystem.createFileSystem(AbstractFileSystem.java:152)
	at org.apache.hadoop.fs.AbstractFileSystem.get(AbstractFileSystem.java:240)
	at org.apache.hadoop.fs.FileContext$2.run(FileContext.java:332)
	at org.apache.hadoop.fs.FileContext$2.run(FileContext.java:329)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:415)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1548)
	at org.apache.hadoop.fs.FileContext.getAbstractFileSystem(FileContext.java:329)
	at org.apache.hadoop.fs.FileContext.getFileContext(FileContext.java:443)
	at org.apache.hadoop.fs.FileContext.getFileContext(FileContext.java:423)
	at org.kitesdk.data.spi.filesystem.FileSystemMetadataProvider.writeDescriptor(FileSystemMetadataProvider.java:543)
```

rdblue: Using the FileContext API, is the rename call still atomic, or does it result in a delete and a rename in the FileSystem API?

ben-roling: Using FileContext the rename w/ overwrite is atomic for HDFS.  It is a bit confusing to trace through the code as there is a call to AbstractFileSystem.rename() which has a non-atomic default implementation.  For HDFS this is overridden in org.apache.hadoop.fs.Hdfs, which contains an atomic implementation.

noslowerdna:
Also note that using FileContext would break S3A filesystem support (until [HADOOP-11262](https://issues.apache.org/jira/browse/HADOOP-11262))

PR #352, opened 2015-04-15
CDK-615: Permissions issue writing to a partitioned Hive external dataset

joey:
* Added a test of `load()` methods of `DatasetWriterCacheLoader` and
  `IncrementalDatasetWriterCacheLoader`.
* Moved the `writer.initialize()` call to happen before the
  `listener.partitionAdded()` call in `DatasetWriterCacheLoader#load`
  and `IncrementalDatasetWriterCacheLoader#load`.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/PartitionedDatasetWriter.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/PartitionedDatasetWriter.java
@@ -15,6 +15,7 @@
  */
 package org.kitesdk.data.spi.filesystem;
 
+import com.google.common.annotations.VisibleForTesting;
 import java.util.Map;
 import org.kitesdk.data.DatasetDescriptor;
 import org.kitesdk.data.DatasetWriter;
@@ -186,7 +187,8 @@ abstract class PartitionedDatasetWriter<E, W extends FileSystemWriter<E>> extend
         .toString();
   }
 
-  private static class DatasetWriterCacheLoader<E> extends
+  @VisibleForTesting
+  static class DatasetWriterCacheLoader<E> extends
     CacheLoader<StorageKey, FileSystemWriter<E>> {
 
     private final FileSystemView<E> view;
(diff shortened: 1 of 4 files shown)


joey:
I've been thinking more about this and my fix here may not be correct. This works for file system datasets and for external Hive datasets, but for managed Hive datasets the directories should probably be created by Hive.

I'm thinking the correct solution is that the partition listener should always be called before the directory is created. The external Hive metadata provider would create it on the Kite side and the managed one would just let the Hive metastore create it.

Thoughts?

joey:
I pushed a new commit to the branch that reverses the logic and makes partition listeners responsible for deciding if Kite or an external system should create the partition directories. This makes Kite work more consistently with both managed and external Hive tables and fixes the original permissions issue.

There is probably still follow-up work on how to deal with the permissions and ownership of files themselves. I think as it stands, you can't drop a Hive managed dataset using the Hive DDL statements as the files are probably not owned by Hive but by the user that ran Kite.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/PartitionedDatasetWriter.java
```diff
@@ -256,6 +262,9 @@ public IncrementalDatasetWriterCacheLoader(FileSystemView<E> view) {
             dataset.getNamespace(), dataset.getName(), partition.toString());
       }
 
+      // initialize the writer after calling the listener
+      // this lets the listner decide if and how to create the
```
rdblue: Typo: listner

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/PartitionedDatasetWriter.java
```diff
@@ -216,6 +218,9 @@ public DatasetWriterCacheLoader(FileSystemView<E> view) {
...
```
rdblue: Typo: Listner

rdblue:
+1

PR #353, opened 2015-04-15
CDK-986. Allow partitioning on optional fields

tomwhite:


rdblue:
+1

joey:
Do we also have to define how the various partitioners handle a null value? For example, if I'm hash partitioned on a field what is the hash of a null value? Is it randomly assigned a partition or does it always go to a specific bucket?

The same kind of questions come up for other partitioners.

rdblue:
I think the intent is to fail with NPE when partition values are null. We don't expect them to actually be null, but we don't want to be so strict about schemas.

tomwhite:
That's right - if a partition value is actually null you'll get a NPE.

tomwhite:
Merged in https://github.com/kite-sdk/kite/commit/3738dffc751cc8b26fceb0d4ce8445f23848bf57

PR #354, opened 2015-04-15
CDK-989: HttpMetricsMorphlineTest assumes only ConnectionException 

chelin:
The documentation of URLConnection.getInputStream() is that it may throw IOException. Which derived IOException which are thrown though is somewhat platform dependent. For some of the platforms a closed socket will instead throw SocketException.


whoschek:
Thanks for the fix!

PR #357, opened 2015-04-16
CDK-843: Add replace to rewrite partitions

rdblue:
This is a work-in-progress. This adds a replace method, similar to merge, that replaces a partition. The use case is to copy data out of a dataset with MR and then replace the original partitions with the rewritten partition, which compacts the data.

The first commit adds `PartitionView` as discussed on the mailing list, and adds `getCoveringPartitions` to the `View` API. The second commit rewrites merge and adds replace in terms of those partition views.

### DIFF
--- a/kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/CompactCommand.java
+++ b/kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/CompactCommand.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2013 Cloudera Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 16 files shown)


rdblue:
@tomwhite or @joey, I think this is ready to review if you have a chance. The last couple commits will change slightly when #359 is merged, but the bulk of this can be reviewed. Just ignore changes to TransformTask. Thanks!

tomwhite:
This looks good to me. +1

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/CompactCommand.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.IOException;
+import java.util.List;
+import org.apache.crunch.PipelineResult;
+import org.kitesdk.data.View;
+import org.kitesdk.tools.CompactionTask;
+import org.slf4j.Logger;
+
+import static org.apache.avro.generic.GenericData.Record;
+
+@Parameters(commandDescription="Compact all or part of a dataset")
```
joey: Does this command handle the case where there is a writer writing to the same partitions that are being compacted?

rdblue: No, and good point. This is an unsafe compaction. I'll add a note to the description that this is unsafe so users know.

PR #358, opened 2015-04-21
CDK-462: Added static block to add Oozie Action Config to Configuration ...

mkwhitacre:
...if present.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DefaultConfiguration.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DefaultConfiguration.java
@@ -17,6 +17,7 @@
 package org.kitesdk.data.spi;
 
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.Path;
 
 /**
  * Manages a default Hadoop {@link Configuration}.
@@ -33,7 +34,19 @@ import org.apache.hadoop.conf.Configuration;
 public class DefaultConfiguration {
 
   // initialize the default configuration from the environment
-  private static Configuration conf = new Configuration();
+  private static Configuration conf;
+
+  static{
+    //read system property for Oozie Action Configuration XML
+    String oozieActionXml = System.getProperty("oozie.action.conf.xml");
+    if(oozieActionXml != null && oozieActionXml.length() > 0) {
+      Configuration.addDefaultResource(new Path("file:///", oozieActionXml).toString());
+    }
+
+    conf = new Configuration();
+  }
+
+
 
   /**
    * Get a copy of the default Hadoop {@link Configuration}.


kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DefaultConfiguration.java
```diff
@@ -33,7 +34,19 @@
...
 
   // initialize the default configuration from the environment
-  private static Configuration conf = new Configuration();
+  private static Configuration conf;
+
+  static{
+    //read system property for Oozie Action Configuration XML
+    String oozieActionXml = System.getProperty("oozie.action.conf.xml");
+    if(oozieActionXml != null && oozieActionXml.length() > 0) {
+      Configuration.addDefaultResource(new Path("file:///", oozieActionXml).toString());
```
tomwhite: I think that default resources are file paths that are loaded from the classpath, i.e. they shouldn't have a file:// schema.

tomwhite:
This looks good. Is it possible to add a test, or at least describe how you tested it?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DefaultConfiguration.java
mkwhitacre: So this code I borrowed from Oozie's Hive2Main[1].  Do default resources behave differently than the config.addResource(...)

[1] - https://github.com/apache/oozie/blob/5e507124ec5fc06e7723f7b6c589afd09143c513/sharelib/hive2/src/main/java/org/apache/oozie/action/hadoop/Hive2Main.java#L74

mkwhitacre:
Since the code is in a static block it is kind of hard to test unit test wise.  I could change it around such that the default config is lazily loaded once and check that there.  So something like:

```
public static Configuration get() {
  if(baseConfig == null){
      //do everything in the static block
  } 
return new Configuration(baseConfig);
}

It would make unit testing a lot easier.  However it would make the class not thread-safe.  Regardless I'll document how I tried this on a cluster with Oozie.  I assume trying with Oozie 4.0 is ok or would you expect 3.x and 4.1?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DefaultConfiguration.java
tomwhite: Yes, I think they do, since default resources are not interpreted as Paths, but Strings: https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/conf/Configuration.java#L2605-L2619

tomwhite:
Manual testing with Oozie 4.0 is fine for this.

rdblue:
@mkwhitacre: I'd like to get this in for 1.1.0, which will hopefully be soon. Is there anything else we need to do or do you think this is ready to be merged? Any comments @tomwhite?

tomwhite:
I was just looking for confirmation that this has been tested (manually is fine) on a cluster. If so, then I'm +1.

rdblue:
Merged as https://github.com/kite-sdk/kite/commit/b60b57f07a03dbbfb344c1afcb14010874a552c4. Thanks @mkwhitacre and @tomwhite!

mkwhitacre:
Sorry, I haven't gotten a chance to test on cluster yet.  I'll still try to get it done in the next week.

PR #359, opened 2015-04-22
CDK-898: Use MRPipeline instead of MemPipeline

rdblue:
This is based on #303, but fixes the logging configuration so that a link to the MR job is given when running on a cluster and task errors are printed to the console when running locally.


rdblue:
The last commit is a work-around for [CRUNCH-512](https://issues.apache.org/jira/browse/CRUNCH-512).

tomwhite:
It looks like the last commit breaks tests. The other changes look good to me though.

rdblue:
Fixed the tests. Changes to the configuration in the TransformTask were leaking into other test cases. The solution was to make a copy if it needs to make changes.

rdblue:
Merged. Thanks for the review, Tom!

PR #360, opened 2015-04-25
CDK-972: Add PartitionView

rdblue:
This was refactored out of #357.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/View.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/View.java
@@ -128,4 +128,28 @@ public interface View<E> {
    * @since 0.16.0
    */
   public URI getUri();
+
+  /**
+   * Returns an {@code Iterable} of non-overlapping {@link View} objects that
+   * partition the underlying {@link Dataset} and cover this {@code View}.
+   *
+   * The returned {@code View} objects are implementation-specific, but should
+   * represent reasonable partitions of the underlying {@code Dataset} based on
+   * its layout.
+   *
+   * The data contained by the union of each {@code View} in the
+   * {@code Iterable} must be a super-set of this {@code View}.
+   *
+   * Note that partitions are actual partitions under which data is stored.
+   * Implementations should omit any {@code View} that is empty.
+   *
+   * @return
+   *      An Iterable of the {@code View} that cover this {@code View}.
+   * @throws IllegalStateException
+   *      If the underlying {@code Dataset} is not partitioned.
+   * @throws UnsupportedOperationException
+   *      If the implementation does not support covering partitions.
+   * @since 1.1.0
+   */
+  public Iterable<PartitionView<E>> getCoveringPartitions();
 }
--- a/kite-data/kite-data-flume/pom.xml
+++ b/kite-data/kite-data-flume/pom.xml
@@ -105,6 +105,12 @@
       <version>1.0.1-SNAPSHOT</version>
     </dependency>
     <dependency>
+      <groupId>org.kitesdk</groupId>
+      <artifactId>${artifact.hadoop-deps}</artifactId>
+      <type>pom</type>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <scope>test</scope>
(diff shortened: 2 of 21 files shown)


kite-data/kite-data-core/src/main/java/org/kitesdk/data/View.java
```diff
@@ -128,4 +128,28 @@
...
+   *
+   * @return
+   *      An Iterable of the {@code View} that cover this {@code View}.
+   * @throws IllegalStateException
+   *      If the underlying {@code Dataset} is not partitioned.
+   * @throws UnsupportedOperationException
+   *      If the implementation does not support covering partitions.
+   * @since 1.1.0
+   */
+  public Iterable<PartitionView<E>> getCoveringPartitions();
```
rbrush: Understanding that we want to grant flexibility to the implementations, but I'm wondering if we can strengthen the guarantees of the returned partitions a bit. For instance, if I have a year/month/day partition, getting the covering partitions for view:...example?year=2015&month=5 could do any of the following per this specification:

* A single-element iterable for the View containing the entire year of 2015 
* A single-element iterable view for just May of 2015 
* A 31-element iterable for each day in May of 2015

At the very least I think we should guarantee getCoveringPartitions() will return no views that are disjoint from itself. For instance, getCoveringPartitions() for year=2015&month=5 won't return the year=2015&month=6 partition, since that is disjoint from the partition we're invoking this on. (This seems obvious but isn't disallowed according to this spec.)

Perhaps we could guarantee that getCoveringPartition returns the minimal number of non-overlapping partitions that contain some data in the given view? This should make it easier to write tooling in the future to, say, efficiently archive off old data.

rdblue: I think you're right that this should require that each view returned overlaps part of the original view and could have data in it. That's definitely an oversight.

As for what should be returned by implementations, I'm not sure what the right thing to do is. The use case I implemented this for needs to operate on the smallest partition level to avoid removing partitions that are peers of partitions that are being replaced (the 31-element iterable in your example). We could change that, but I'm not sure which is more correct.

One consideration: the root will be the minimal partition in a lot of cases, when you have no partition constraints. In that case, it's misleading to think you can simply move the directories because there are .metadata and .signals folders that would be lost. So maybe the right thing to do is to return all of the full paths as partition views and allow other FS code to analyze what's on disk to merge those into work units. But then, that puts a lot of work to get efficient operations on the user so I'm not too happy either way.

Thoughts?

rbrush: Returning the most granular level of partitions like you describe seems fine to me...it's predictable and easy to understand. I think we could do things in addition to this (and probably outside the scope of this change) to make it easier for management tooling. 

Just a thought, but some possible methods that could help do this on the PartitionView interface: 

* A getParentPartition() method returning a parent, if it exists. This would make it easier to "walk up" from the more granular views.
* A getPartitionKeys(), which returns a map of the keys defining a partition: ("year" -> "2015", "month" -> "5", "day" -> 22), which allows tooling to find the partition they want to work with.

This is a bit hypothetical...something like this might be best to add in a subsequent JIRA once we have a concrete design for arching data. This is assuming you don't mind adding new methods to the PartitionView interface after it has been published. (One of those technically non-passive but doubtful to break users interface additions.)

rdblue: I agree, there might be more methods we should provide on PartitionView, but let's hold off until we have a concrete use case. Tom was +1 on #357, which still includes this so I'd like to merge this soon. Is that alright?

rbrush: +1. Agreed that we shouldn't add methods on a hypothetical, so we can revisit when we have something concrete.

rdblue:
@tomwhite could you take a look at this? This separated the PartitionView addition from the other PR and added tests. That ended up requiring some changes to the directory iterator code that I think should be looked at. Thanks!

kite-data/kite-data-flume/pom.xml
```diff
@@ -105,6 +105,12 @@
       <version>1.0.1-SNAPSHOT</version>
     </dependency>
     <dependency>
+      <groupId>org.kitesdk</groupId>
```
tomwhite: Not sure why kite-data-flume dependencies needs to be changed here.

rdblue: Flume relies on StorageKey (SPI), which now exposes a Path. That was causing the tests to fail because the Hadoop dependency wasn't present. That was because this is provided everywhere and isn't ever a transitive dep.

tomwhite: OK, thanks for the explanation.

tomwhite:
Looks good, +1.

rdblue:
Thanks, Tom!

PR #364, opened 2015-05-13
CDK-476 oozie UriHandler for Kite URIs

bbrownz:
An Oozie URIHandler for KiteURIs. This allows KiteURIs to be used for data input events. Particularly when chaining together workflows based on the availability of an upstream dataset (a populated view in this case).

Uses the signal ready concept implemented in CDK-451 as an alternative to the traditional "_SUCCESS" file.

### DIFF
--- a/kite-data/kite-data-oozie/pom.xml
+++ b/kite-data/kite-data-oozie/pom.xml
@@ -1,0 +1,194 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+~ Copyright 2015 Cloudera Inc.
+~
+~ Licensed under the Apache License, Version 2.0 (the "License");
+~ you may not use this file except in compliance with the License.
+~ You may obtain a copy of the License at
+~
+~ http://www.apache.org/licenses/LICENSE-2.0
+~
+~ Unless required by applicable law or agreed to in writing, software
+~ distributed under the License is distributed on an "AS IS" BASIS,
+~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+~ See the License for the specific language governing permissions and
+~ limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>kite-data-oozie</artifactId>
+
+  <parent>
+    <groupId>org.kitesdk</groupId>
+    <artifactId>kite-data</artifactId>
+    <version>1.0.1-SNAPSHOT</version>
+  </parent>
+
+  <name>Kite Data Oozie Module</name>
+  <description>
+    The Kite Data Oozie module provides Oozie support for working with Kite datasets.
+  </description>
+
+  <properties>
+    <vers.oozie4>4.1.0</vers.oozie4>
+  </properties>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-source-plugin</artifactId>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-javadoc-plugin</artifactId>
+        <configuration>
+          <excludePackageNames>org.kitesdk.data.oozie</excludePackageNames>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.rat</groupId>
+        <artifactId>apache-rat-plugin</artifactId>
+      </plugin>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>findbugs-maven-plugin</artifactId>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-antrun-plugin</artifactId>
+        <executions>
+          <execution>
+            <phase>compile</phase>
+            <goals>
+              <goal>run</goal>
+            </goals>
+            <configuration>
+              <tasks>
+                <echo message="Create empty javadoc JAR to satisfy Maven central" />
+                <mkdir dir="target/apidocs" />
+              </tasks>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+
+  <reporting>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-project-info-reports-plugin</artifactId>
+        <reportSets>
+          <reportSet>
+            <inherited>false</inherited>
+            <reports>
+              <report>index</report>
+              <report>summary</report>
+              <report>dependency-info</report>
+              <report>dependencies</report>
+            </reports>
+          </reportSet>
+        </reportSets>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-javadoc-plugin</artifactId>
+      </plugin>
+    </plugins>
+  </reporting>
+
+  <dependencies>
+    <!-- Kite -->
+    <dependency>
+      <groupId>org.kitesdk</groupId>
+      <artifactId>kite-data-core</artifactId>
+      <version>${project.parent.version}</version>
+    </dependency>
+
+    <!-- Hadoop -->
+    <dependency>
+      <groupId>org.kitesdk</groupId>
+      <artifactId>${artifact.hadoop-deps}</artifactId>
+      <type>pom</type>
+      <scope>provided</scope>
+      <exclusions>
+        <exclusion>
+          <!-- conflicts with jetty included with HBase -->
+          <groupId>tomcat</groupId>
+          <artifactId>jasper-runtime</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.kitesdk</groupId>
+      <artifactId>kite-hadoop-compatibility</artifactId>
+    </dependency>
+
+    <!-- custom URI handlers require Oozie 4.x -->
+    <dependency>
+      <groupId>org.apache.oozie</groupId>
+      <artifactId>oozie-core</artifactId>
+      <version>${vers.oozie4}</version>
+    </dependency>
+
+    <!-- Misc -->
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
+      <scope>provided</scope>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>annotations</artifactId>
+      <scope>provided</scope>
+    </dependency>
+
+    <!-- Test Dependencies -->
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.kitesdk</groupId>
+      <artifactId>kite-data-core</artifactId>
+      <version>${project.parent.version}</version>
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.kitesdk</groupId>
+      <artifactId>${artifact.hadoop-test-deps}</artifactId>
+      <type>pom</type>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <profiles>
+    <profile>
+      <id>cdh5</id>
+      <activation>
+        <property>
+          <name>hadoop.profile</name>
+          <value>cdh5</value>
+        </property>
+      </activation>
+      <properties>
+        <vers.oozie4>${vers.oozie}</vers.oozie4>
+      </properties>
+    </profile>
+  </profiles>
+
+</project>
--- a/kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
+++ b/kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
@@ -1,0 +1,157 @@
+/**
+ * Copyright 2015 Cloudera Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kitesdk.data.oozie;
+
+import java.net.URI;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.oozie.ErrorCode;
+import org.apache.oozie.XException;
+import org.apache.oozie.action.hadoop.LauncherURIHandler;
+import org.apache.oozie.dependency.URIHandler;
+import org.apache.oozie.dependency.URIHandlerException;
+import org.apache.oozie.service.HadoopAccessorException;
+import org.apache.oozie.service.URIHandlerService;
+import org.apache.oozie.util.XLog;
+import org.kitesdk.data.DatasetException;
+import org.kitesdk.data.DatasetNotFoundException;
+import org.kitesdk.data.Datasets;
+import org.kitesdk.data.Signalable;
+import org.kitesdk.data.URIBuilder;
+import org.kitesdk.data.View;
+
+/**
+ * A Kite URI handler that works with {@link Signalable} views.
+ *
+ * To be considered as {@link #exists(URI, Configuration, String) existing} the
+ * view must have been signaled as ready.
+ */
+public class KiteURIHandler implements URIHandler {
+
+  private static final XLog LOG = XLog.getLog(KiteURIHandler.class);
+
+  private Set<String> supportedSchemes;
+  private List<Class<?>> classesToShip;
+
+  @Override
+  public void init(final Configuration conf) {
+    supportedSchemes = new HashSet<String>();
+    final String[] schemes = conf.getStrings(
+        URIHandlerService.URI_HANDLER_SUPPORTED_SCHEMES_PREFIX
+            + this.getClass().getSimpleName()
+            + URIHandlerService.URI_HANDLER_SUPPORTED_SCHEMES_SUFFIX, "view", "dataset");
+    supportedSchemes.addAll(Arrays.asList(schemes));
+    classesToShip = new KiteLauncherURIHandler().getClassesForLauncher();
+  }
+
+  @Override
+  public Set<String> getSupportedSchemes() {
+    return supportedSchemes;
+  }
+
+  @Override
+  public Class<? extends LauncherURIHandler> getLauncherURIHandlerClass() {
+    return KiteLauncherURIHandler.class;
+  }
+
+  @Override
+  public List<Class<?>> getClassesForLauncher() {
+    return classesToShip;
+  }
+
+  @Override
+  public DependencyType getDependencyType(final URI uri)
+      throws URIHandlerException {
+    return DependencyType.PULL;
+  }
+
+  @Override
+  public void registerForNotification(final URI uri, final Configuration conf,
+      final String user, final String actionID) throws URIHandlerException {
+    throw new UnsupportedOperationException(
+        "Notifications are not supported for " + uri);
+  }
+
+  @Override
+  public boolean unregisterFromNotification(final URI uri, final String actionID) {
+    throw new UnsupportedOperationException(
+        "Notifications are not supported for " + uri);
+  }
+
+  @Override
+  public Context getContext(final URI uri, final Configuration conf,
+      final String user) throws URIHandlerException {
+    return null;
+  }
+
+  @SuppressWarnings("rawtypes")
+  @Override
+  public boolean exists(final URI uri, final Context context) throws URIHandlerException {
+    try {
+      View<GenericRecord> view = Datasets.load(uri);
+      if(view instanceof Signalable) {
+        return ((Signalable)view).isReady();
+      }
+    } catch (IllegalArgumentException ex) {
+      LOG.error("the URI " + uri + " was not a view or dataset");
+    } catch (DatasetNotFoundException ex) {
+      LOG.error("the dataset for the URI "+ uri +" did not actually exist");
+    } catch (DatasetException e) {
+      throw new HadoopAccessorException(ErrorCode.E0902, e);
+    }
+    // didn't meet all the requirements for a URI/view that we want to use with oozie
+    return false;
+  }
+
+  @Override
+  public boolean exists(final URI uri, final Configuration conf, final String user)
+      throws URIHandlerException {
+    // currently does not handle access limitations between the oozie user and
+    // a potential dataset owner
+    return exists(uri, null);
+  }
+
+  @Override
+  public String getURIWithDoneFlag(final String uri, final String doneFlag)
+      throws URIHandlerException {
+    return uri;
+  }
+
+  @Override
+  public void validate(final String uri) throws URIHandlerException {
+    try {
+      URI uriParsed = URI.create(uri);
+      String scheme = uriParsed.getScheme();
+      if(!(URIBuilder.VIEW_SCHEME.equals(scheme) || URIBuilder.DATASET_SCHEME.equals(scheme))) {
+        LOG.error("Unexpected scheme: view, uri was "+ uri);
+        XException xException = new XException(ErrorCode.E0904, scheme, uri);
+        throw new URIHandlerException(xException);
+      }
+    } catch (IllegalArgumentException iae) {
+      XException xException = new XException(ErrorCode.E0906, iae);
+      throw new URIHandlerException(xException);
+    }
+  }
+
+  @Override
+  public void destroy() {
+  }
+
+}
--- a/kite-data/kite-data-oozie/src/test/java/org/kitesdk/data/oozie/TestKiteURIHandler.java
+++ b/kite-data/kite-data-oozie/src/test/java/org/kitesdk/data/oozie/TestKiteURIHandler.java
@@ -1,0 +1,392 @@
+/**
+ * Copyright 2015 Cloudera Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kitesdk.data.oozie;
+
+import org.kitesdk.data.PartitionView;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import org.apache.avro.Schema;
+import org.apache.avro.SchemaBuilder;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.oozie.ErrorCode;
+import org.apache.oozie.XException;
+import org.apache.oozie.dependency.URIHandlerException;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.kitesdk.data.Dataset;
+import org.kitesdk.data.DatasetDescriptor;
+import org.kitesdk.data.DatasetReader;
+import org.kitesdk.data.DatasetWriter;
+import org.kitesdk.data.Formats;
+import org.kitesdk.data.MiniDFSTest;
+import org.kitesdk.data.PartitionStrategy;
+import org.kitesdk.data.Signalable;
+import org.kitesdk.data.RefinableView;
+import org.kitesdk.data.View;
+import org.kitesdk.data.oozie.KiteURIHandler;
+import org.kitesdk.data.spi.DatasetRepository;
+import org.kitesdk.data.spi.OptionBuilder;
+import org.kitesdk.data.spi.Registration;
+import org.kitesdk.data.spi.URIPattern;
+import org.kitesdk.data.spi.filesystem.FileSystemDatasetRepository;
+
+@RunWith(Parameterized.class)
+public class TestKiteURIHandler extends MiniDFSTest {
+
+  protected static final String NAMESPACE = "ns1";
+  protected static final String NAME = "provider_test1";
+
+  @Parameterized.Parameters
+  public static Collection<Object[]> data() {
+    Object[][] data = new Object[][] {
+        { false },  // default to local FS
+        { true } }; // default to distributed FS
+    return Arrays.asList(data);
+  }
+
+  // whether this should use the DFS provided by MiniDFSTest
+  protected boolean distributed;
+
+  protected Configuration conf;
+  protected DatasetDescriptor testDescriptor;
+  protected FileSystem fs;
+
+  public TestKiteURIHandler(boolean distributed) {
+    this.distributed = distributed;
+  }
+
+  public DatasetRepository newRepo() {
+    return new FileSystemDatasetRepository.Builder()
+        .configuration(conf)
+        .rootDirectory(URI.create("target/data"))
+        .build();
+  }
+
+  @After
+  public void removeDataPath() throws IOException {
+    fs.delete(new Path("target/data"), true);
+  }
+
+  @Before
+  public void setUp() throws IOException, URISyntaxException {
+    this.conf = (distributed ?
+        MiniDFSTest.getConfiguration() :
+        new Configuration());
+
+    this.fs = FileSystem.get(conf);
+
+    this.testDescriptor = new DatasetDescriptor.Builder()
+        .format(Formats.AVRO)
+        .schema(SchemaBuilder.record("Event").fields()
+            .requiredLong("timestamp")
+            .requiredString("message")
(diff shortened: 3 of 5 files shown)


kite-data/kite-data-oozie/src/test/java/org/kitesdk/data/oozie/TestKiteURIHandler.java
```diff
@@ -0,0 +1,378 @@
...
+  public void supportedSchemes() {
+    uriHandler.init(new Configuration());
+    Set<String> scheme = new HashSet<String>();
+    scheme.add("view");
+    scheme.add("dataset");
+    Assert.assertEquals(scheme, uriHandler.getSupportedSchemes());
+  }
+
+  @Test (expected = UnsupportedOperationException.class)
+  public void registerNotificatiosn() throws URISyntaxException, URIHandlerException {
```
rdblue: Typo: Notificatiosn

kite-data/kite-data-oozie/src/test/java/org/kitesdk/data/oozie/TestKiteURIHandler.java
```diff
@@ -0,0 +1,378 @@
...
+  public void checkURIDoesNotExist() throws URIHandlerException, IOException{
+    DatasetRepository repository = newRepo();
+    Dataset<GenericRecord> dataset = repository.create("data","notreadymailbox", testDescriptor);
+
+    RefinableView<GenericRecord> view = dataset.with("message", "hello");
+
+    Assert.assertFalse(uriHandler.exists(view.getUri(), null));
+  }
+
+  @SuppressWarnings("rawtypes")
```
rdblue: Rather than suppressing the warning, why not just add the type?

bbrownz:
Created a PR to kite-examples showing a use case for the URIHandler, https://github.com/kite-sdk/kite-examples/pull/26

kite-data/kite-data-oozie/src/test/java/org/kitesdk/data/oozie/TestKiteURIHandler.java
bbrownz: Good catch 35ec9e

kite-data/kite-data-oozie/src/test/java/org/kitesdk/data/oozie/TestKiteURIHandler.java
bbrownz: Updated in 35ec9e5

rdblue:
Hi @bbrownz, sorry it is taking me so long to get to this. I'm not very familiar with Oozie, but I talked with someone today that was interested and will probably have a look. Otherwise I'll follow up on this next week. Thanks for being patient!

bbrownz:
No worries! The biggest questions I had posting this in were
*we build the project, since not all the profiles made a lot of sense with the URIHandler being in oozie 4, which we only get under the CDH5 profile now. Currently the oozie version is just always locked in.
*If we should support deleting the views through the LauncherURIHandler, that felt like something we wouldn't want for the first iteration though

kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+    // didn't meet all the requirements for a URI/view that we want to use with oozie
+    return false;
+  }
+
+  @Override
+  public boolean exists(final URI uri, final Configuration conf, final String user)
+      throws URIHandlerException {
+    // currently does not handle access limitations between the oozie user and
+    // a potential dataset owner
```
rdblue: Maybe this should attempt to read the signal files and catch any IOExceptions that are thrown. Then this would have at least basic support.

kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
```diff
@@ -0,0 +1,156 @@
...
+
+  @Override
+  public String getURIWithDoneFlag(final String uri, final String doneFlag)
+      throws URIHandlerException {
+    return uri;
+  }
+
+  @Override
+  public void validate(final String uri) throws URIHandlerException {
+    if(uri!=null){
```
rdblue: If `uri == null`, this doesn't throw an exception. Is that okay?

rdblue:
I have a couple of minor comments, but this looks good to me overall. Have you tested this with a cluster?

rdblue:
@ben-roling, could you review this as well?

bbrownz:
> Have you tested this with a cluster?

So far we've tested this on 5.2 and 5.3 with quickstart VMs via the example (kite-sdk/kite-examples#26). We're in progress doing some testing now with one of our dev clusters that has Kerberos enabled.

rdblue:
Sounds good! I'd like someone more familiar with Oozie to take a look, but I'm happy with this going in. Since Ben is a committer, I think it makes sense to get his +1 before merging. Thanks for working on this, @bbrownz!

kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
bbrownz: Good question, this method is a bit ill defined. For example the [FSURIHandler](https://github.com/apache/oozie/blob/release-4.0.0/core/src/main/java/org/apache/oozie/dependency/FSURIHandler.java#L116-L117) doesn't actually do any validation. While the HCatURIHandler doesn't do a null check and would likely through a NPE rather than a URIHandlerException.

@ben-roling what do you think on this one?

kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
bbrownz: Were you thinking along the route of reading the signal files directly (so we'd be able to flex the configuration and user used for the FileSystem access)? That seems like it would bind us very directly to the current implementation, although it would work.

rdblue: No, I meant try to do the operation like you do here, but if it fails with a permissions-related IOException, return false instead of propagating the exception.

kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
ben-roling: I don't think the null handling is terribly important.  The URIHandler interface doesn't express an expectation of how it should be handled.  Oozie is the only thing that I would expect to be calling the validate() method and I suspect it won't ever be calling with a null URI anyway.

I suppose if it were up to me though I would throw or allow an exception to happen on a null URI.  Perhaps the simplest thing to do is just remove the null check and allow an uncaught NPE to happen (which would occur on URI.create(uri) on line 138).

ben-roling:
@bbrownz works with me and as such I had already reviewed most of this prior to the PR being posted.  I'll take another quick pass through and then I can formally document my +1.

kite-data/kite-data-oozie/pom.xml
```diff
@@ -0,0 +1,172 @@
...
+        <artifactId>maven-antrun-plugin</artifactId>
+        <executions>
+          <execution>
+            <phase>compile</phase>
+            <goals>
+              <goal>run</goal>
+            </goals>
+            <configuration>
+              <tasks>
+                <echo message="Create empty javadoc JAR to satisfy Maven central" />
```
ben-roling: This is a bit confusing to me.  Why is this ant task necessary?  I presume this is copy/paste from one of the other sub-projects.  It looks like it is in kite-data-hbase, kite-data-hive, kite-data-s3, and kite-hadoop-compatibility.

tomwhite: It's because the module does not have any public classes so there's no javadoc for it. This task generates an empty javadoc JAR, which is needed for Maven central.

kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
bbrownz: Ah, I follow, thats an interesting trade off. 

I think returning false would make the most sense if we were approaching it from the perspective that the (oozie process) user without permissions shouldn't know that the dataset exists or not. Propagating the exception through helps to reveal the issue that the permissions are blocking the operation they were hoping would work.

I think propagating the exception makes the most sense in that alert a consumer that their permission setup isn't going to work with the URI handler as is. If we were to create documentation around the usage we would definitely want to note that case (outside of the example project I didn't create any, and perhaps should have).

kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
bbrownz: Updated to remove the null check which would allow the NPE to propagate

6ace20d2

kite-data/kite-data-oozie/pom.xml
ben-roling: Ok, that explanation makes sense in the context of the other projects but this project does have public classes and with this ant task removed it does still generate a javadoc JAR.

I think maybe this task just needs to be removed from this POM.  Alternatively we can update the package excludes filter for the javadoc plugin (defined in the parent POM) to exclude org.kitesdk.data.oozie directly indicating we choose not to publish javadoc.  I don't expect anyone is really going to be reading the javadoc as it's not really a Kite consumer oriented project.

@tomwhite or @rdblue - do you have an opinion about whether or not we should publish the javadoc for these classes?

tomwhite: I don't think we should publish javadoc for this module as it doesn't contain classes that users will program to in their programs. So we should update the package excludes filters and leave the ant task as it is in this patch.

kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  @Override
+  public boolean unregisterFromNotification(final URI uri, final String actionID) {
+    throw new UnsupportedOperationException(
+        "Notifications are not supported for " + uri);
+  }
+
+  @Override
+  public Context getContext(final URI uri, final Configuration conf,
```
ben-roling: We probably should be implementing this method.  The method is a mechanism to optimize a series of calls to exist().  Its usage can be seen in [CoordELFucntions.coord_futureRange_sync](https://github.com/apache/oozie/blob/release-4.0.0/core/src/main/java/org/apache/oozie/coord/CoordELFunctions.java#L313) and [coord_latestRange_sync](https://github.com/apache/oozie/blob/release-4.0.0/core/src/main/java/org/apache/oozie/coord/CoordELFunctions.java#L1044).

The coord_latestRange_sync method is used to look for the [latest complete instance (or instances) of a dataset](https://oozie.apache.org/docs/4.0.0/CoordinatorFunctionalSpec.html#a6.6.6._coord:latestint_n_EL_Function_for_Synchronous_Datasets) when coord:future() is used in the coordinator definition.  Oozie will loop backwards from the latest nominal time all the way back to the initial instance of the dataset looking for an instance that "exists" (is complete).  In the worst case you have a dataset that has been around for along time and the latest instance is very old relative to the frequency of the dataset.  For example, if the dataset is declared with a frequency of once per day with initial instance declared as 1 year ago and the only instance that is complete is actually one year ago then latest(0) will have to loop through 365 URIs checking exists() before returning that instance.

The coord_futureRange_sync method supports [coord:future()](https://oozie.apache.org/docs/4.0.0/CoordinatorFunctionalSpec.html#a6.6.7._coord:futureint_n_int_limit_EL_Function_for_Synchronous_Datasets) and loops forward looking for the desired number of compete instances.

FSURIHandler [uses getContext() to cache the Filesystem instance](https://github.com/apache/oozie/blob/release-4.0.0/core/src/main/java/org/apache/oozie/dependency/FSURIHandler.java#L82).

HCatURIHandler [caches the HCatClient](https://github.com/apache/oozie/blob/release-4.0.0/core/src/main/java/org/apache/oozie/dependency/HCatURIHandler.java#L143).

I think we could cache the View instance so we don't have as heavy of a hit on each exists() check to load up all the Dataset metadata.

Separately from talking about this specific method, I had talked to @bbrownz about the idea that perhaps we might want to follow up later with a JIRA to introduce some other caching for optimization.  In an environment with lots of Oozie coordinators using Kite datasets and this URI handler, the Oozie server could be making very frequent calls to exists().  Having each call hit HDFS and/or the Hive metastore to load dataset metadata (on Datasets.load()) might take a toll that could be decreased with some more caching of the Dataset instances.

There is also the possibility that we could follow up later with an attempt to support a push model (DependencyType.PUSH) instead of a polling model for Hive backed datasets.

In any case, I think what is here works and is probably worth merging in as-is and following up with the optimizations in separate later JIRAs.  Perhaps we implement the getContext() piece now though as it is the simplest although it only comes into play in certain specific use cases (using coord:latest() or coord:future() - neither of which I expect my organization will actually use anytime soon as our use cases prefer coord:current()).

kite-data/kite-data-oozie/pom.xml
rdblue: I agree with Tom. Most of the code here is implementing a public API documented in Oozie, right?

kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
rdblue: @bbrownz or @ben-roling, could you file the follow-up JIRA issues before we merge this one? That way we keep track.

kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
rdblue: Sounds fine to me.

ben-roling: There is precedent for catching IOException and rethrowing as HadoopAccessorException in [FSURIHandler](https://github.com/apache/oozie/blob/release-4.0.0/core/src/main/java/org/apache/oozie/dependency/FSURIHandler.java#L91).  It seems reasonable to me to be consistent with that.

kite-data/kite-data-oozie/pom.xml
ben-roling: Correct and I agree it would be fine not to publish javadoc.  These classes will not be consumed by "users" - just by Oozie.

kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
ben-roling: Sure, I'll log the JIRA and post a link here in a minute.

ben-roling: https://issues.cloudera.org/browse/CDK-1007

ben-roling:
I think there are just 3 things left with this:
* exclude the package on the javadoc plugin config
* possibly change the IOException handling for exists()
* implement getContext()

Once @bbrownz implements the trivial javadoc plugin config change I am +1.

I don't care that much about the IOException handling thing.  It could be changed to rethrow as HadoopAccessorException or left alone and either way would be fine by me.

With CDK-1007 I have provided a sub-task to defer the getContext() change.  I think we could merge this without it.

kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
bbrownz: I added a catch in the exists method for a DatasetException to re-throw that exception as a HadoopAccessorException. I'd considered the more specific DatasetIOException that IOExceptions are wrapped in but this seemed like it would cover our bases a little better. It also occurred to me that HadoopAccessorException might be less appropriate in the future (for example if we added a signal implementation that was HBase specific) but that felt like I had started to nitpick myself. I'm open to future opinions though :smile: 

a310ebf

kite-data/kite-data-oozie/pom.xml
bbrownz: Added the exclusion to the javadoc plugin 32186c3

bbrownz:
Addressed @ben-roling's comments from above. I think it's best to leave the implementation of the getContext() method to the optimization in the CDK-1007 issue when the usage would really come into play.

I everything looks good here I can squash this down into a single change before the merge.

rdblue:
Squash away!

bbrownz:
Rebased and squashed, should be ready after the travis build goes through.

kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * A Kite URI handler that works with {@link Signalable} views.
+ *
+ * To be considered as {@link #exists(URI, Configuration, String) existing} the
+ * view must have been signaled as ready.
+ */
+public class KiteURIHandler implements URIHandler {
+
+  private static final Logger LOG = LoggerFactory.getLogger(KiteURIHandler.class);
```
rkanter: in Oozie, we use our XLog class for logging; the methods should be the same though.  For example:

`private static XLog LOG = XLog.getLog(KiteURIHandler.class);`

rkanter:
Hi, I'm one of the Oozie devs.  I took a look at this, and it looks fine to me.  I made a minor comment about logging.  That said, the Yahoo! Oozie devs are more familiar with the URIHandler code; they're the ones who generalized it and added support for HCat URIs.

What do you think about including this directly in Oozie?  That would make things easier for users because they wouldn't have to add jars into the Oozie server.  And we can include the example Coordinator too.  I know that Ben was working on OOZIE-1829, though I guess that's no longer needed for now.  Anyway, I think putting this in Oozie would be a great integration story, and I can also ask one of the Yahoo! Oozie devs to review the code.

rdblue:
Thanks for taking a look @rkanter!

@bbrownz and @ben-roling, we should definitely decide if we want this to go in Oozie or Kite. I'd like to avoid adding it here only to deprecate it in favor of an Oozie version in a release or two. My vote is to get this into Oozie, since that seems like the right place for it to live, as with Flume and the DatasetSink. What do you guys think?

ben-roling:
Thanks for the feedback @rkanter!

At first glance I thought it would be nice to have this included in Oozie itself as well but @bbrownz and I talked through it more and I'm not so sure.

The primary concern we had was with regard to the speed at which we could get changes pushed through the system.  For example, if a consumer needs a fix to Kite they need to wait for a Kite release, an Oozie release after that, and then finally a distro release before they finally have it in their environment.

If the handler is only included in Kite things can work much more quickly.  Once the new Kite version is released they just deploy it and it is done with no concern for when the next Oozie or distro release might happen.

All this said, there is nothing we could do to stop the Oozie devs from incorporating Kite URI support directly into their project if they wanted to.  If they did, it could be difficult for consumers to override the Kite version if they needed to pick up a newer Kite version more quickly than they could get a new Oozie version.

Assuming we continue down a path of keeping the packaging separate from Oozie their is also the possibility we could do a documentation patch to Oozie such that Oozie doc makes mention of the Kite integration, making the functionality easier for users to find.

Thoughts?

kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
bbrownz: good call, updated to XLog here: c743837

rkanter:
That's a good point.  We don't have a regular release schedule for Oozie; it's just sort of whenever someone feels like doing it.  And we don't have frequent releases.  If Kite wants to iterate faster on it, then it would make sense to put it in Kite.  If we do that, I do like the idea of mentioning it in the Oozie docs.

rdblue:
How much do we expect this component to evolve? It seems like we should be designing it so that it touches Kite's public API -- which I think is how it currently works -- and doesn't need to change much since it is just doing change detection based on URI and the signalling API. I know we've identified some improvements and opened issues for them, but shouldn't a short-term goal be to get this to the point where we don't need to change it often?

If we could just update the version of Kite on a cluster and have Oozie pick up the latest, then I think that would be preferable.

ben-roling:
I'm not expecting this component specifically to change that much.  I was thinking more about Kite core changing.  Of course this component doesn't use the full surface area of the Kite API so there are only certain types of changes to Kite core that could affect the Oozie integration.  The change would have to be something that affects Datasets.load() or Signalable.isReady() - maybe a bug fix or an enhancement or optimization of some sort.

 > If we could just update the version of Kite on a cluster and have Oozie pick up the latest, then I think that would be preferable.

Per your suggestion, we could take a hybrid approach where we include the KiteURIHandler class in Kite itself and do an Oozie patch that adds configuration to pick it up.

More specifically, the Oozie patch would:
* update oozie-default.xml to add the new URI schema to URIHandler class mapping
* include a default Kite version in the Oozie dependencies so things work out of the box
* include a configuration mechanism to point to an alternate Kite installation path as a mechanism to pick up a newer/different Kite install

I don't off the top of my head know exactly how that would be implemented on the Oozie side but we could look into it if that direction is desirable.

rkanter:
That is another option.  The code could live in Kite and Oozie just pulls it in like any other 3rd-party maven dependency during the build.  We already parameterize a number of dependencies in the pom, so we could just add a "kite.version".  Then, anyone who builds Oozie can easily switch versions by changing the pom, and anyone who already has Oozie can simply swap out the kite jar with a newer version.  That leaves all of the code in Kite, while still making it work in Oozie out-of-the-box.

ben-roling:
My hope is to avoid a solution where when a consumer wants a new Kite version before an Oozie release it requires rebuilding Oozie or modifying the files provided by the existing Oozie deployment in their cluster.

@rkanter - the way you describe sounds to me like you are suggesting a solution that would require deleting JARs from the existing Oozie deployment to get the new Kite version to go into effect.

My thought was slightly different in that we would add some mechanism to Oozie where the Kite dependencies live in some specific directory and an environment variable or configuration key of some sort can be set to point to a non-default directory to override the Kite install.

Ideally the mechanism would be such that in the CDH world a parcel could be used to apply a new Kite version to Oozie.  If deleting files out of the original Oozie deployment is required I don't think a parcel would be able to do the job.

rkanter:
In CDH, all those jars would be symlinked.  So, if you update Kite and restart Oozie, Oozie should automatically pick the new jars up.  I was talking more from an Apache release perspective, with a tarball install.  i.e. You have Oozie 4.1.0, and you want to update the Kite jars, you can just add the new Kite jars and redeploy the Oozie server.  You'd have to do this anyway if Oozie didn't ship with any Kite jars, right?

kite-data/kite-data-oozie/pom.xml
```diff
@@ -0,0 +1,175 @@
...
+    <dependency>
+      <groupId>org.kitesdk</groupId>
+      <artifactId>kite-hadoop-compatibility</artifactId>
+    </dependency>
+
+    <!-- custom URI handlers require Oozie 4.x -->
+    <dependency>
+      <groupId>org.apache.oozie</groupId>
+      <artifactId>oozie-core</artifactId>
+      <version>4.0.0-cdh${cdh5.version}</version>
```
rdblue: I just spotted this because it caused CI test failures. This should be the upstream Apache version of Oozie instead of the CDH one. We test against CDH in the cdh5 profile, so you can override the version to be 4.1.0-cdh5.4.2 there. (Looks like it is based on 4.1.0, which is why the test failed in the first place.)

bbrownz: Ah, yeah, it looks like when I rebased and squashed I pulled in the update for the CDH5 profile to 5.4.2 (which included an update in the CDH5 profile to go to the 4.1.0 version of oozie).

This project will still build under the default and CDH4 profiles where "vers.oozie" would be 3.3.2 based (before the URI handler). So just "vers.oozie" wouldn't work. 

Just to make sure I'm in sync, are you suggesting I just set this to 4.1.0 and drop the CDH portion?

rdblue: I think the version in the cdh5 profile should be 4.1.0-cdh5.4.2. I don't mind what we use for upstream, but feel free to update it if that works better for you.

ben-roling:
> In CDH, all those jars would be symlinked. So, if you update Kite and restart Oozie, Oozie should automatically pick the new jars up.

Yeah, but I'm not sure you follow what I mean.  What would your "update Kite" steps be?  Let's pretend we add kite-data as a dependency of Oozie and in CDH5.5 I get this new version of Oozie installed into my environment as part of the standard CDH5.5 parcel.  Now kite-data-1.1.jar is in `/opt/cloudera/parcels/CDH-5.5.0/lib/oozie/libserver`.  Now, kite-data-1.1.1 comes out with a critical (to me) bug fix and I want to get it into my environment.  Cloudera hasn't deemed this fix important enough so there is no CDH-5.5.1 parcel for me to pick it up from.  How do I incorporate it?

Sure, I suppose I could go manually muck with the `/opt/cloudera/parcels/CDH-5.5.0/lib/oozie/libserver` directory on my system but I don't want to do that.  What I was thinking was I wanted to create my own parcel that I can install with a fix.  I guess I could create my own `CDH-5.5.0-myfix` parcel that repackages the whole CDH5.5 distribution but I wasn't wanting to go there.  I want to be able to create a parcel that contains only the new Kite JARs and install that as an addition to the system on top of the CDH-5.5.0 parcel.  I can do that but the problem is there is no way that I am aware (via a parcel) of for me to take the original Kite JARs off the Oozie classpath without repackaging all of Oozie or all of CDH.  Even if I wanted (and I don't) to take the route of just repackaging all of Oozie I'm not yet familiar enough with the parcel system to know how I could do that either.

> I was talking more from an Apache release perspective, with a tarball install. i.e. You have Oozie 4.1.0, and you want to update the Kite jars, you can just add the new Kite jars and redeploy the Oozie server. You'd have to do this anyway if Oozie didn't ship with any Kite jars, right?

Maybe it depends on what you mean by "redeploy the Oozie server".  If Oozie didn't ship any changes related to this feature what a user would do to use the feature is change their oozie-site.xml, add the necessary JARs to /var/lib/oozie, and restart Oozie.  They don't have to touch any of the content that was shipped by Oozie or in any way rebuild Oozie.  I wouldn't consider that "redeploying the Oozie server" but you could be using a different definition of the word "deploy".

In the model I just described If there was a new release of Kite you needed you would delete the Kite-related JARs from /var/lib/oozie, put the new ones in, and restart Oozie.

The direction I was trying to suggest is such that instead of dumping the Kite JARs in /var/lib/oozie you would dump them wherever you like (e.g. /my/kite/jar/path) and then update some Oozie config to add /my/kite/jar/path to the classpath.  When you want a new version of Kite you can either delete the old jars from /my/kite/jar/path and add the new ones or you can create a new /my/kite/jar/path/v2 folder and update the Oozie config to point to the new location.

With this model Oozie could ship default Kite JARs in some folder embedded in Oozie and add that directory to the classpath unless the configuration is overridden to point to another directory.

Am I starting to make more sense?

rkanter:
>Cloudera hasn't deemed this fix important enough so there is no CDH-5.5.1 parcel for me to pick it up from. How do I incorporate it?  
>What I was thinking was I wanted to create my own parcel that I can install with a fix. I guess I could create my own CDH-5.5.0-myfix parcel that repackages the whole CDH5.5 distribution but I wasn't wanting to go there. I want to be able to create a parcel that contains only the new Kite JARs and install that as an addition to the system on top of the CDH-5.5.0 parcel.

I'm not familiar with the release process for Kite, but I was I don't know if we want to go down the rabbit-hole of specifically helping users doing custom things like patching and building their own version of Kite (unless that's common for Kite?).  And as far as I know, when you build a Parcel, you have to include every component, not just the updated jars; so I don't think you can do that.

> Now kite-data-1.1.jar is in /opt/cloudera/parcels/CDH-5.5.0/lib/oozie/libserver. Now, kite-data-1.1.1 comes out with a critical (to me) bug fix and I want to get it into my environment.
For CDH components, `/opt/cloudera/parcels/CDH-5.5.0/lib/oozie/libserver` doesn't have version numbers in the jar names.  Instead, it uses symlinks to unversioned jars in that component's dir, which are actually symlinks that point to versioned jars.  So, `/opt/cloudera/parcels/CDH-5.5.0/lib/oozie/libserver` would have `kite-data.jar` which points to `../../../lib/kite/lib/kite-data.jar` (or something similar), which in-turn points to `kite-data-1.1.jar`.  When you upgrade the Kite parcel, it should have a new unversioned symlink that points to `kite-data-1.1.1.jar`.

> Maybe it depends on what you mean by "redeploy the Oozie server". If Oozie didn't ship any changes related to this feature what a user would do to use the feature is change their oozie-site.xml, add the necessary JARs to /var/lib/oozie, and restart Oozie.
Redeploying the Oozie server is only for tarball installs, which is what is done upstream.  And `/var/lib/oozie` only works for CDH package or parcel installs.  Upstream Oozie doesn't have a binary distribution.  When you want to install that Oozie, you download the source tarball and compile it.  There's then a setup script that you have to run, which lets you add additional jars to the WAR file before deploying it in Tomcat.  In upstream Oozie, if you want to change or add additional jars, you have to re-run the setup script to redeploy Oozie.  In CDH packages/parcels, we do some magic stuff with exploded WAR files where you don't have to do this, and /var/lib/oozie is symlinked into the deployed WAR "file" in Tomcat.

> The direction I was trying to suggest is such that instead of dumping the Kite JARs in /var/lib/oozie you would dump them wherever you like (e.g. /my/kite/jar/path) and then update some Oozie config to add /my/kite/jar/path to the classpath.
Because of Tomcat, Oozie doesn't have lots of control over manipulating it's classpath.  Currently, we've only made classes configurable in oozie-site (e.g. Here's the list of classes to load on startup, here's the list of URIHandler classes, etc), and Oozie just assumes their already on the classpath.  And for that to happen, the jars have to be in the right place as I described in my previous paragraph.

Another option is to only include the Kite jar with the CDH version of Oozie, and require users of the upstream version of Oozie download it manually.  Mixing CDH versions is unsupported, so this shouldn't be a problem; and again, the unversioned symlinks help here anyway.  We currently do this with some of the optional dependencies in Oozie, so there's precedent for that.

rkanter:
By the way, I'll be on vacation until Tuesday with limited access, so you probably won't get another reply from me until then.

kite-data/kite-data-oozie/pom.xml
bbrownz: I think this commit should get us closer to what we want dbf34f9091

It's still awkward in that the module always needs to build with at least oozie 4.x, and the CDH4 and default profiles both build with oozie 3.3.2. To help get around this I added a vers.oozie4 property to the data-oozie module, and extended the cdh5 profile to set vers.oozie4 to use the CDH version under that profile. This way we should still be using the CDH5 version when testing under that profile and the oozie 4 version under the other two profiles.

ben-roling:
Thanks for all the feedback and extra information @rkanter. I guess my thoughts were probably mostly rooted in my organization's current usage of kite in a mentality where kite is not a distribution-provided library. We're still mostly on CDH4 where it seems to me kite is not distribution-provided. Now that I am looking more closely it seems that kite is a distribution-provided library in at least recent CDH5 parcels. For example, I have a Quickstart VM with a CDH5.3.3 parcel on it at the moment and I see Kite is included inside it.

In our non-distribution-provided usage so far we have just picked up Kite versions as Kite project releases them (independent from CDH releases). I have not even thought to look to see the relationship between Kite releases and CDH releases. For example, what is the average time between a Kite release and a CDH parcel release that follows it? If those generally happen in tandem then clearly my whole thinking was wrong as there would always be a CDH parcel release I could pick up to get the change.

To be clear, my goal wasn't to try to support use cases of users running forks of Kite. I was targeting use cases of official Kite releases that hadn't gotten into the distribution yet. Like I indicated in my previous paragraph, perhaps the idea of Kite releases that aren't in the distribution yet is fallacy? Maybe such things don't really exist or only exist for very short periods of time.

I should also note that I am not yet as familiar as I'd like to be with the parcel concepts in CDH. I understand there is a core CDH parcel and then there are some other additive parcels you can install on top (e.g. LZO support).

I was talking to a colleague yesterday and he pointed me to [this thread](https://groups.google.com/a/cloudera.org/forum/#!topic/scm-users/Altw2esOVx0) that helped me understand things a bit better. Parcels are expected to be disjoint and there isn't really a reliable mechanism to override something in one parcel with something else in another parcel.

I guess the idea I had before was kind of like the idea of having a parcel that overrides the kite from the CDH parcel but not exactly. I was thinking more like it was something that adds an additional Kite version to the system and then allows an Oozie configuration option that allows the administrator to use it just for Oozie. That said, the idea doesn't really make sense when you think of kite as a distribution-provided dependency across the platform (not just in the context of Oozie). If I wanted Oozie to use a new Kite version I would most likely want the whole platform to use the new Kite version. At that point what I would want would be more directly like the idea of having a parcel that overrides the CDH provided cross-platform Kite library, which is something the current parcel system doesn't support.

In the end I guess it appears to me the correct way to get a new Kite version in the context of CDH is to get a whole new CDH parcel and presumably just to wait for Cloudera to release one. Again, I haven't yet checked the relationship between Kite releases and unsolicited CDH parcel releases but if there is too big of a gap there then of course the consumer would try to coerce Cloudera to perform and unscheduled parcel packaging. It's now my understanding that it isn't terribly uncommon for Cloudera to prepare an unscheduled parcel for a customer that needs some fix before the fix is scheduled to go in a regular CDH update release.

So, I’m giving up on the idea of the configuration option being added to Oozie.  We still have a choice to make about how to incorporate these changes into Oozie:
* All the code goes into Kite and we add nothing except a documentation reference to Oozie
* The kite-oozie module stays in Kite and we add Kite dependencies and ooze-default.xml update to Oozie
* The kite-oozie module moves to oozie (likely named oozie-kite instead since now it is an oozie module instead of a kite module) such that there is nothing related to the URIHandler in Kite itself at all except maybe a documentation reference

I presume @rdblue and @rkanter you both prefer the 2nd or 3rd option over the 1st?  Which do you prefer of the 2nd and 3rd options?

rdblue:
@ben-roling, I'm inclined to keep this in Kite for now and update Oozie documentation. It looks like Oozie releases are rare and we'll be able to progress much more quickly with Kite's more regular release timeline. That said, if we end up not changing anything in this for a few months, I think we should revisit the issue and move it to Oozie because we will have confirmed that it won't change much as long as it uses Kite's public API.

As far as CDH releases go, we do keep up with Kite features in those releases. CDH 5.4.x has Kite 1.0.0. Since Kite is both a client library, you can either use it in your code (whatever version you like) or use it through components that aren't libraries. Kite shipped with CDH is primarily for components like Flume that use Kite internally, but also for customers that prefer to pick up changes more slowly and not have to update their software with the upstream project.

Next steps: let's get this merged! Is there anything that needs to be fixed before adding this to Kite?

bbrownz:
I don't believe there is anything left to fix. I pushed up a new rebased version. Should be ready to go if everything looks good on the build.

ben-roling:
Thanks for the feedback @rdblue - it will be nice to finally have a released form of this functionality!

We can certainly revisit the possibility of moving this over to the Oozie project in the future.

rkanter:
I've created https://issues.apache.org/jira/browse/OOZIE-2269 to add the docs to Oozie.

Given that CDH releases ship with a single version of Kite (and the user isn't expected to update it without updating CDH), I'll probably look into having the kite-oozie jars in Oozie automatically in just CDH; that will save those users some extra steps.

rdblue:
@rkanter, that sounds good to me.

rdblue:
@bbrownz: local tests are failing because the column projection API was merged before this. Sorry! Could you update?

bbrownz:
Sure, pushed up a quick rebase against master. Local tests looked good on the default profile.

bbrownz:
It looks like the CDH5 profile build failed on the kite-data-flume module with the native thread error. It's likely unrelated but it might be worth kicking off another travis build just to be sure.

> java.lang.OutOfMemoryError: unable to create new native thread

ben-roling:
@rkanter - thanks for logging OOZIE-2269 - I think @bbrownz or I could probably get something for you to review on that.

If you can also get the update into CDH to include kite-oozie automatically that sounds nice.

rdblue:
Merged as https://github.com/kite-sdk/kite/commit/d540ead0

It's great to add support for Oozie, thanks for making it happen @bbrownz and @ben-roling!

PR #365, opened 2015-05-14
CDK-1000: Kite/SparkSQL integration

dgreco:
I integrated what I've already done here: https://github.com/dgreco/spark-kite into the Spark trunk

### DIFF
--- a/kite-app-parent/cdh5/pom.xml
+++ b/kite-app-parent/cdh5/pom.xml
@@ -67,15 +67,20 @@
   </scm>
 
   <properties>
+    <!-- Scala versions -->
+    <scala.base.version>2.10</scala.base.version>
+    <scala.version>2.10.5</scala.version>
+    <scalatest.version>2.2.4</scalatest.version>
+
     <!-- dependency version properties -->
     <kite.version>1.0.1-SNAPSHOT</kite.version>
-    <kite.avro.version>1.7.5-cdh5.1.0</kite.avro.version>
-    <kite.crunch.version>0.10.0-cdh5.1.0</kite.crunch.version>
+    <kite.avro.version>1.7.6-cdh5.2.0</kite.avro.version>
+    <kite.crunch.version>0.11.0-cdh5.2.0</kite.crunch.version>
     <kite.findbugs.version>2.0.1</kite.findbugs.version>
-    <kite.hive.version>0.12.0-cdh5.1.0</kite.hive.version>
-    <kite.oozie.version>4.0.0-cdh5.1.0</kite.oozie.version>
-    <kite.spark.version>1.0.0-cdh5.1.0</kite.spark.version>
-    <kite.parquet.version>1.2.5-cdh5.1.0</kite.parquet.version>
+    <kite.hive.version>0.13.1-cdh5.2.0</kite.hive.version>
+    <kite.oozie.version>4.0.0-cdh5.2.0</kite.oozie.version>
+    <kite.spark.version>1.3.0-cdh5.4.0</kite.spark.version>
+    <kite.parquet.version>2.1.0-cdh5.2.0</kite.parquet.version>
 
     <!-- other properties to help apps -->
     <hadoop.guava.version>11.0.2</hadoop.guava.version>
@@ -401,6 +406,32 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.apache.spark</groupId>
+      <artifactId>spark-sql_${scala.base.version}</artifactId>
+      <scope>provided</scope>
+      <optional>true</optional>
+      <exclusions>
+        <!-- Exclude Spark's Hadoop dependencies -->
+        <exclusion>
+          <artifactId>hadoop-annotations</artifactId>
+          <groupId>org.apache.hadoop</groupId>
+        </exclusion>
+        <exclusion>
+          <artifactId>hadoop-auth</artifactId>
+          <groupId>org.apache.hadoop</groupId>
+        </exclusion>
+        <exclusion>
+          <artifactId>hadoop-client</artifactId>
+          <groupId>org.apache.hadoop</groupId>
+        </exclusion>
+        <!-- Exclude conflicting servlet-api -->
+        <exclusion>
+          <artifactId>servlet-api</artifactId>
+          <groupId>javax.servlet</groupId>
+        </exclusion>
+      </exclusions>
+    </dependency>
 
   </dependencies>
 
@@ -408,7 +439,7 @@
     <dependencies>
       <dependency>
         <groupId>org.apache.spark</groupId>
-        <artifactId>spark-core_2.10</artifactId>
+        <artifactId>spark-core_${scala.base.version}</artifactId>
         <version>${kite.spark.version}</version>
         <scope>provided</scope>
         <optional>true</optional>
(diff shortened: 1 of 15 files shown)


kite-app-parent/cdh5/pom.xml
```diff
@@ -67,15 +67,20 @@
...
+    <kite.crunch.version>0.11.0-cdh5.2.0</kite.crunch.version>
     <kite.findbugs.version>2.0.1</kite.findbugs.version>
-    <kite.hive.version>0.12.0-cdh5.1.0</kite.hive.version>
-    <kite.oozie.version>4.0.0-cdh5.1.0</kite.oozie.version>
-    <kite.spark.version>1.0.0-cdh5.1.0</kite.spark.version>
-    <kite.parquet.version>1.2.5-cdh5.1.0</kite.parquet.version>
+    <kite.hive.version>0.13.1-cdh5.2.0</kite.hive.version>
+    <kite.oozie.version>4.0.0-cdh5.2.0</kite.oozie.version>
+    <kite.spark.version>1.3.0-cdh5.4.0</kite.spark.version>
+    <kite.parquet.version>2.1.0-cdh5.2.0</kite.parquet.version>
```
tomwhite: There's a mixture of CDH5 versions here - go to 5.4.1?

kite-data/kite-data-spark/pom.xml
```diff
@@ -52,9 +71,24 @@
         <groupId>org.apache.rat</groupId>
         <artifactId>apache-rat-plugin</artifactId>
       </plugin>
+<!-- Disabled becaues it conflicts with Scala classes
```
tomwhite: This should be fixed before commit.

kite-data/kite-data-spark/src/main/scala/com/databricks/spark/avro/SchemaSupport.scala
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.databricks.spark.avro
+
+import org.apache.avro.{Schema, SchemaBuilder}
+import org.apache.spark.sql.SQLContext
+import org.apache.spark.sql.types.{DataType, StructType}
+
+trait SchemaSupport {
```
tomwhite: Is this not available as a dependency?

kite-data/kite-data-spark/src/main/scala/org/kitesdk/data/spark/KiteDatasetSaver.scala
```diff
@@ -0,0 +1,53 @@
...
+    val converter = createConverter(schema, "topLevelRecord")
+    rows.map(x => (converter(x).asInstanceOf[Record], null))
+  }
+
+  def saveAsKiteDataset(dataFrame: DataFrame, uri: URI, format: Format, compressionType: CompressionType, partitionStrategy: Option[PartitionStrategy]): Dataset[Record] = {
+    assert(URIBuilder.DATASET_SCHEME == uri.getScheme, s"Not a dataset or view URI: $uri" + "")
+    val job = Job.getInstance()
+
+    //TODO this is an hack, the partitioning strategy only works with primitive type not nullable, if the schema has been built using reflection from a case class for example, the primitive tyoe are nullable in the avro schema
+    //breaking the partition strategy, so in case the partition strategy is defined I change nullable to false, it seems working.
```
tomwhite: https://issues.cloudera.org/browse/CDK-986 should mean this hack is no longer needed.

kite-data/kite-data-spark/src/main/scala/com/databricks/spark/avro/SchemaSupport.scala
dgreco: Unfortunately not, the databricks spark-avro project provides mechanisms for reading/writing data frames form/to avro files. It internally uses classes that convert data frame schemas to AVRO schemas and vice versa. Unfortunately either they are private or they are private at the package level. I wanted to reuse as much I could what already done in that project this is why I had to introduce that hacks and to put something under this package. It's ugly I understand but I couldn't do otherwise.

tomwhite:
Thanks for submitting this PR @dgreco. At a high-level, I wonder if it would be more useful to work with RDDs, since data frames are pretty new. I was imagining a SparkDatasets class (like CrunchDatasets):
```java
RDD asRDD(View view)
RDD asRDD(URI datasetUri)
void saveAsDataset(RDD rdd, View view)
void saveAsDataset(RDD rdd, URI datasetUri)
```
Dataframes could be added to an interface like this too.

What do you think?

kite-data/kite-data-spark/src/main/scala/org/kitesdk/data/spark/KiteDatasetSaver.scala
```diff
@@ -0,0 +1,53 @@
...
+import org.kitesdk.data._
+import org.kitesdk.data.mapreduce.DatasetKeyOutputFormat
+
+object KiteDatasetSaver extends SchemaSupport {
+  private def rowsToAvro(rows: Iterator[Row], schema: StructType): Iterator[(Record, Null)] = {
+    val converter = createConverter(schema, "topLevelRecord")
+    rows.map(x => (converter(x).asInstanceOf[Record], null))
+  }
+
+  def saveAsKiteDataset(dataFrame: DataFrame, uri: URI, format: Format, compressionType: CompressionType, partitionStrategy: Option[PartitionStrategy]): Dataset[Record] = {
```
tomwhite: Please keep lines under 80 chars in width.

kite-app-parent/cdh5/pom.xml
dgreco: I only updated Spark to the latest version available keeping the rest at the lower level possible. I could upgrade everything if you want.

kite-data/kite-data-spark/src/main/scala/org/kitesdk/data/spark/KiteDatasetSaver.scala
dgreco: I'll get rid off that hack

dgreco:
Working with Dataframes is nice because they contain schema information as well like kite datasets. 
They are new but are improving very quickly and they are becoming really fundamental in Spark.
Working directly with Dataframes is really complex especially when you have to prepare data for machine learning. In my experience people tend to prefer data frames to plain rdds. 

BTW, I'm adding also java tests, I could generate another pull requests with your suggestions and the additional Java tests.

kite-app-parent/cdh5/pom.xml
tomwhite: Got it. I would update to the latest. Is there any downside to doing that, @rdblue?

tomwhite:
That makes a lot of sense. Do you think you could change the public API to be more like the CrunchDatasets one? E.g.

```java
DataFrame asDataFrame(View view)
DataFrame asDataFrame(URI datasetUri)
void saveAsDataset(DataFrame dataFrame, View view)
void saveAsDataset(DataFrame dataFrame, URI datasetUri)
```

dgreco:
Yes I can, nice suggestions. Those are nice for the Java side. From the Scala side I thought it was nice to  mimic what's currently available on SparkSQL: sqlContext.parquetFile("people.parquet") (saveAsKiteDataset) just adding an implicit I could extend sqlContext for saving DF as kite data sets.
So what I can do is to follow your convention on the Java side and to add proper implicits on the Scala side, what do you think?

tomwhite:
Yes, that sounds like a good direction.

dgreco:
Ok, I'll do it. Stay tuned for the next PR.

dgreco:
I have a question, if I implement this:

void saveAsDataset(DataFrame dataFrame, URI datasetUri)

how can I pass the DatasetDescriptor? in my implementation I'm passing format, compression, etc. and I build inside the needed descriptor for creating the dataset.

kite-app-parent/cdh5/pom.xml
rdblue: I think we should update everything now that 5.4.x is out. We should be tracking the releases for CDH or other distros we support more closely. But I think this is a separate issue that we should do before merging this one.

rdblue:
Another API to consider (before you make all the changes) is [the API that spark-avro uses](https://github.com/databricks/spark-avro#scala-api). That looks reusable and seems to be what databricks intends for the standard.

kite-app-parent/cdh5/pom.xml
dgreco: I did already this exercise on a separate branch if you want I could do it this upgrade. If you are not rushing, I have a limited bandwidth.

dgreco:
@rdblue exactly, this is what I did with the current APIs, I tried to follow the spark avro ones.

dgreco:
@rdblue and the spark-avro apis follow the approach of spark sql

kite-data/kite-data-spark/src/main/scala/org/kitesdk/data/spark/KiteDatasetReader.scala
```diff
@@ -0,0 +1,33 @@
...
+ */
+package org.kitesdk.data.spark
+
+import org.apache.hadoop.mapreduce.Job
+import org.apache.spark.sql.{DataFrame, SQLContext}
+import org.kitesdk.data.Dataset
+
+object KiteDatasetReader {
+
+  def readAsDataFrame(sqlContext: SQLContext,
```
tomwhite: Naming: should this be KiteDatasetLoader and loadAsDataFrame, for symmetry with KiteDatasetSaver/saveAsKiteDataset.

PR #366, opened 2015-05-19
CDK-1000: Kite/SparkSQL integration

dgreco:


kite-data/kite-data-spark/src/main/scala/org/kitesdk/data/spark/KiteDatasetReader.scala
```diff
@@ -0,0 +1,33 @@
...
+ */
+package org.kitesdk.data.spark
+
+import org.apache.hadoop.mapreduce.Job
+import org.apache.spark.sql.{DataFrame, SQLContext}
+import org.kitesdk.data.Dataset
+
+object KiteDatasetReader {
+
+  def readAsDataFrame(sqlContext: SQLContext,
```
tomwhite: Rename KiteDatasetLoader/loadAsDataFrame for symmetry with KiteDatasetSaver/saveAsKiteDataset.

kite-data/kite-data-spark/src/main/scala/org/kitesdk/data/spark/KiteDatasetSaver.scala
```diff
@@ -0,0 +1,166 @@
...
+    /*
+    TODO this is an hack, the partitioning strategy only works with primitive
+    type not nullable, if the schema has been built using reflection from a
+    case class for example, the primitive types are nullable in the avro schema
+    breaking the partition strategy, so in case the partition strategy is defined
+    I change nullable to false, it seems working.
+    */
+    val schema = partitionStrategy.fold(dataFrame.schema)(_ =>
+      StructType(dataFrame.schema.iterator.map(field =>
+        if (field.nullable) field.copy(nullable = false) else field).toArray))
```
tomwhite: Did https://issues.cloudera.org/browse/CDK-986 not remove the need for this?

kite-data/kite-data-spark/src/main/scala/org/kitesdk/data/spark/KiteDatasetSaver.scala
```diff
@@ -0,0 +1,166 @@
...
+    val converter = createConverter(schema, "topLevelRecord")
+    rows.map(x => (converter(x).asInstanceOf[Record], null))
+  }
+
+  private def saveAsKiteDataset(dataFrame: DataFrame,
+                                uri: URI,
+                                format: Format,
+                                compressionType: CompressionType,
+                                partitionStrategy: Option[PartitionStrategy],
+                                columnMapping: Option[ColumnMapping],
```
tomwhite: These parameters should be wrapped up in a DatasetDescriptor. Also we should think about whether we need a DatasetDescriptor at all, or whether it is optional, so that existing datasets can be used. CrunchDatasets appends data, and has a handleExisting check (see DatasetTarget).

kite-data/kite-data-spark/src/main/scala/org/kitesdk/data/spark/KiteDatasetSaver.scala
```diff
@@ -0,0 +1,166 @@
...
+                                job: Option[Job] = None): Dataset[Record] = {
```
tomwhite: We shouldn't need to expose Job here.

kite-data/kite-data-spark/src/test/scala/org/kitesdk/data/spark/SparkKiteSpec.scala
```diff
@@ -0,0 +1,305 @@
...
+
+  "Spark" must {
+    "be able to create a SchemaRDD/Dataframe from a kite parquet dataset" in {
+
+      cleanup()
+
+      val conf = new SparkConf().
+        setAppName("spark-kite-spec-test").
+        set("spark.serializer", "org.apache.spark.serializer.KryoSerializer").
+        setMaster("local[16]")
```
tomwhite: Can this just be `local`?

kite-data/kite-data-spark/src/test/scala/org/kitesdk/data/spark/SparkKiteSpec.scala
```diff
@@ -0,0 +1,305 @@
...
+
+      sparkContext.stop()
+
+      cleanup()
+
+    }
+  }
+
+  "Spark" must {
+    "be able to create a SchemaRDD/Dataframe from a kite avro dataset" in {
```
tomwhite: This test is basically the same as the previous one, except it uses Avro - can you factor out into a common method?

kite-data/kite-data-spark/src/test/scala/org/kitesdk/data/spark/SparkKiteSpec.scala
```diff
@@ -0,0 +1,305 @@
...
+        setAppName("spark-kite-spec-test").
+        set("spark.serializer", "org.apache.spark.serializer.KryoSerializer").
+        setMaster("local[16]")
+
+      val sparkContext: SparkContext = new SparkContext(conf)
+
+      val sqlContext = new SQLContext(sparkContext)
+      import sqlContext.implicits._
+
+      val datasetURI = URIBuilder.build(s"repo:file:////${System.getProperty("user.dir")}/tmp", "test", "persons")
```
tomwhite: Create temp paths under target.

tomwhite:
Thanks for the update @dgreco! I added some more review comments. Have you looked at why the Travis tests are failing?

It might be good to get more eyes on this - @joey wrote the original Spark integration so it would be good to have his take on this change.

You don't need to open new PRs for each change BTW. It's best to have one so that the discussion is in one place.

kite-data/kite-data-spark/src/main/scala/org/kitesdk/data/spark/KiteDatasetSaver.scala
dgreco: Nope

kite-data/kite-data-spark/src/main/scala/org/kitesdk/data/spark/KiteDatasetReader.scala
dgreco: Yes, it's better

kite-data/kite-data-spark/src/main/scala/org/kitesdk/data/spark/KiteDatasetSaver.scala
dgreco: I was thinking to pass a descriptor instead, The problem I found is that I have to specify everything in the descriptor including the schema, but the schema is generated internally from the passed data frame.

For writing I'm using the kite output format, so if it's possible to append using the kite output format I could do it as well.

kite-data/kite-data-spark/src/test/scala/org/kitesdk/data/spark/SparkKiteSpec.scala
dgreco: I'll do it.

kite-data/kite-data-spark/src/main/scala/org/kitesdk/data/spark/KiteDatasetSaver.scala
dgreco: I'm passing an Option[Job] so if it's defined I'll use the passed job otherwise I would create a Job instance from the external context. In some tests I need to pass a job, because it's used to pass the configuration associated to the base mini cluster. That parameter is None by default so it's not used normally.

kite-data/kite-data-spark/src/main/scala/org/kitesdk/data/spark/KiteDatasetSaver.scala
dgreco: An alternative could to specify a specific SparkDatasetDescriptor that allows specifying the dataframe which is then used internally for generating the schema, something like this:
```scala
val descriptor = new SparkDatasetDescriptor.Builder().dataFrame(teenagers).format(format).compressionType(CompressionType.Snappy).build()

val dataset = KiteDatasetSaver.saveAsKiteDataset(descriptor, datasetURI)
```
What do you think?

dgreco:
The spark specific tests are passing now, it seems now that some of the tests are failing because of lacking of resources

kite-data/kite-data-spark/src/main/scala/org/kitesdk/data/spark/KiteDatasetSaver.scala
tomwhite: Kite has a slightly different way of passing configuration information, via the DefaultConfiguration. You should be able to use that from tests, and so not pass it here.

kite-data/kite-data-spark/src/main/scala/org/kitesdk/data/spark/KiteDatasetSaver.scala
tomwhite: Yes, this is better.

tomwhite:
One of the test failures is due to multiple Spark contexts running at once. Can you try setting spark.driver.allowMultipleContexts to true?

kite-data/kite-data-spark/src/main/scala/org/kitesdk/data/spark/KiteDatasetSaver.scala
dgreco: ok, good suggestion. I'll fix that.

dgreco:
Tom, I committed other stuff, commits:  8d79a53 and  53aa87e 
I'm still learning how to use PRs, did you check what I did? I changed the apis, I tried to reuse the 
concept of dataset descriptor to be passed to the save method, instead of passing multiple parameters.
I also added a method for appending/overwriting to existing datasets. It looks nicer to me, clean apis.
And less code.

tomwhite:
Thanks for the latest changes @dgreco. It looks like the Spark tests are failing with NPEs:

https://travis-ci.org/kite-sdk/kite/jobs/63672318

Also Flume, possibly due to changes in the library versions:

https://travis-ci.org/kite-sdk/kite/jobs/63672316

Do these pass locally?

dgreco:
Hi Tom,
yes they are passing locally, I'm trying with the same commands that travis is using.
It looks like that travis is exhausting resources.

dgreco:
Hi Tom,
finally I made the tests running on Travis, the problem was related to the spark context management when there are multiple spark tests. I refactored the tests to be sure that only one spark context is running at same moment.

dgreco:
I close this PR since I'm restructuring my git repo

PR #367, opened 2015-05-21
CDK-973: Add View#asSchema and View#asType projection methods

rdblue:
This extends Joey's PR, #346, and fixes the review issues I noted. One of the main issues was that without changing the backing dataset's type class, projection would fail. This adds `asType` projection to take care of that issue.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemViewKeyInputFormat.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemViewKeyInputFormat.java
@@ -20,6 +20,8 @@ import com.google.common.collect.Lists;
 import java.io.IOException;
 import java.util.Iterator;
 import java.util.List;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericData;
 import org.apache.avro.hadoop.io.AvroSerialization;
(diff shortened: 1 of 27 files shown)


rdblue:
@joey, could you review these updates to your original PR? I went ahead and made the changes because I want to make sure this is in the next release.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemViewKeyInputFormat.java
```diff
@@ -71,37 +69,38 @@
...
-            SpecificData.get().getSchema(dataset.getType()).toString());
-      }
-    } else if (Formats.PARQUET.equals(format)) {
-
-      // Use the reader's schema type if provided.
-      if (isSpecific) {
-
-        AvroReadSupport.setAvroReadSchema(conf,
-            SpecificData.get().getSchema(dataset.getType()));
+    if (!DataModelUtil.isGeneric(dataset.getType())) {
```
rdblue: I don't think this should only happen if the data model is not generic. If it is generic, we can still set the data model and schema correctly.

joey:
+1, let me know if you want me to merge this or if you want to handle it.

rdblue:
I'll merge. Thanks for reviewing it!

PR #368, opened 2015-05-21
CDK-976: DatasetKeyInputFormat/DatasetKeyOutputFormat setting job con…

EdwardSkoviak:
…figuration before loading dataset

### DIFF
--- a/kite-data/kite-data-mapreduce/src/main/java/org/kitesdk/data/mapreduce/DatasetKeyOutputFormat.java
+++ b/kite-data/kite-data-mapreduce/src/main/java/org/kitesdk/data/mapreduce/DatasetKeyOutputFormat.java
@@ -47,6 +47,7 @@ import org.kitesdk.data.spi.Constraints;
 import org.kitesdk.data.spi.DataModelUtil;
 import org.kitesdk.data.spi.DatasetRepositories;
 import org.kitesdk.data.spi.DatasetRepository;
+import org.kitesdk.data.spi.DefaultConfiguration;
 import org.kitesdk.data.spi.Mergeable;
 import org.kitesdk.data.spi.PartitionKey;
 import org.kitesdk.data.spi.Registration;
@@ -499,6 +500,8 @@ public class DatasetKeyOutputFormat<E> extends OutputFormat<E, Void> {
 
   @Override
   public OutputCommitter getOutputCommitter(TaskAttemptContext taskAttemptContext) {
+    Configuration conf = Hadoop.TaskAttemptContext.getConfiguration.invoke(taskAttemptContext);
+    DefaultConfiguration.set(conf);
     View<E> view = load(taskAttemptContext);
     return usePerTaskAttemptDatasets(view) ?
         new MergeOutputCommitter<E>() : new NullOutputCommitter();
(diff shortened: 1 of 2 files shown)


kite-data/kite-data-mapreduce/src/main/java/org/kitesdk/data/mapreduce/DatasetKeyOutputFormat.java
```diff
@@ -536,6 +536,7 @@ private static String getTaskAttemptDatasetName(TaskAttemptContext taskContext)
     Class<E> type = getType(jobContext);
 
     String outputUri = conf.get(KITE_OUTPUT_URI);
+    DefaultConfiguration.set(conf);
```
rdblue: Should this be done in `OutputCommitter#setupTask` instead of here? This method could be called more than once, but we want to ensure we set up the environment config just one time per task. The `setupTask` methods seems like a good place, unless you have a reason to do it here?

rdblue:
Thanks @EdwardSkoviak! I just have a couple of comments about where the configuration gets set. Thank you for taking the time to fix this!

rdblue:
The Travis build appears to be failing because it can't find DefaultConfiguration. Maybe you're missing an import?

EdwardSkoviak:
> The Travis build appears to be failing because it can't find DefaultConfiguration. Maybe you're missing an import?

Yea sorry, I just snagged the line I added from our forked code and popped it in. The changes you propose above make sense, I will go ahead and make them.

EdwardSkoviak:
@rdblue I made the proposed changes, let me know if it looks good and I'll squash it into one commit.

rdblue:
@EdwardSkoviak, what testing did you do? Would you mind making sure these changes fix your problem if you haven't already?

EdwardSkoviak:
@rdblue The original commit is a change that we've tested and verified that it solved the issue. I'll reverify with these changes but currently I've got a lot on my plate. I'll try to squeeze in this evening or tomorrow morning.

rdblue:
Thanks! I really appreciate it.

EdwardSkoviak:
@rdblue I'm going to test this out shortly, sorry for the delay.

rdblue:
No problem at all, thanks for taking the time to help!

EdwardSkoviak:
So with the test scenarios I've been running it looks like based on the above changes the fix no longer works. Getting the following[1] which prompted the initial change set. I will try to look at this more this weekend if i get a chance.

[1] https://gist.github.com/EdwardSkoviak/0b3ed6971fd4e4c51624

rdblue:
@EdwardSkoviak, thanks for the trace. Looks like we call load in getOutputCommitter because we need to check which committer to use. That means we need to setup the environment in getOutputCommitter rather than in the committer. So I think the fix is to add it to what is currently line 508, just before calling load. Then revert the changes to both committers. That should do it.

EdwardSkoviak:
@rdblue Went ahead and made that change. Just tested it out, works good. Thanks for the help.

rdblue:
@EdwardSkoviak, I'm closing this because I had to make changes to get tests to pass. I'll monitor the tests and get this in as #371. Thanks for your contribution!

PR #369, opened 2015-05-28
CDK-1004: Update compaction for unpartitioned datasets.

rdblue:


kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemDataset.java
```diff
@@ -399,45 +396,50 @@ public void replace(View<E> update) {
...
-        throw new IllegalArgumentException(
-            "Incompatible PartitionView: " + src.getClass().getName());
+        } else {
+          throw new IllegalArgumentException(
+              "Incompatible PartitionView: " + src.getClass().getName());
+        }
       }
+    } else {
+      deleteAll(); // remove all existing files
+      merge((FileSystemDataset<E>) update.getDataset());
```
tomwhite: How about making this safer by moving the existing files to a hidden temporary directory then deleting after the merge succeeds?

PR #371, opened 2015-05-28
CDK-976: Fix task configuration in MR

rdblue:
This sets the default configuration using the one passed to the InputFormat or OutputFormat when running in a MR task. To avoid this setting leaking into tests, this adds `DefaultConfiguration.init(Configuration)`, which will set the default conf only if it has not already been set via `set` or `init`. (Otherwise, test fail.)

This replaces #368.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DefaultConfiguration.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DefaultConfiguration.java
@@ -34,6 +34,7 @@ import org.apache.hadoop.fs.Path;
 public class DefaultConfiguration {
 
   // initialize the default configuration from the environment
+  private static boolean initDone = false;
   private static Configuration conf;
 
   static{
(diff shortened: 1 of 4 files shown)


rdblue:
@tomwhite, could you take a look at this? I think it is correct to set the default config, but I needed to avoid setting it when not running in MR.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DefaultConfiguration.java
```diff
@@ -66,5 +67,16 @@ public static Configuration get() {
...
+  }
+
+  /**
+   * Initialize the default Hadoop {@link Configuration} if it has not been set.
+   */
+  public static void init(Configuration conf) {
+    if (!initDone) {
+      DefaultConfiguration.conf = conf;
+      DefaultConfiguration.initDone = true;
+    }
```
tomwhite: All of the methods on this class should be `synchronized`. Otherwise this change looks good.

rdblue:
Fixed, thanks Tom!

PR #374, opened 2015-05-28
CDK-1006: Update CDH5 profile to 5.4.2.

rdblue:


rbrush:
Should the version of Spark be upgraded as well?

rdblue:
@rbrush, good catch. I'll update this.

PR #377, opened 2015-06-04
CDK-1011: Support a configurable number of writers per partition when…

mkwhitacre:
… writing to a dataset, along with copying and compaction.

I didn't see any current tests for the CrunchDatasets.partition(...) method or CompactionCommand/Task.  I can add some if necessary.

### DIFF
--- a/kite-data/kite-data-crunch/src/main/java/org/kitesdk/data/crunch/CrunchDatasets.java
+++ b/kite-data/kite-data-crunch/src/main/java/org/kitesdk/data/crunch/CrunchDatasets.java
@@ -191,12 +191,40 @@ public class CrunchDatasets {
   public static <E> PCollection<E> partition(PCollection<E> collection,
                                              View<E> view,
                                              int numWriters) {
+    return partition(collection, view, numWriters, 1);
+  }
+
+  /**
+   * Partitions {@code collection} to be stored efficiently in {@code View}.
+   * <p>
+   * This restructures the parallel collection so that all of the entities that
+   * will be stored in a given partition will be evenly distributed across a specified
+   * {@code numPartitionWriters}.
+   * <p>
+   * If the dataset is not partitioned, then this will structure all of the
+   * entities to produce a number of files equal to {@code numWriters}.
+   *
+   * @param collection a collection of entities
+   * @param view a {@link View} of a dataset to partition the collection for
+   * @param numWriters the number of writers that should be used
+   * @param numPartitionWriters the number of writers data for a single partition will be distributed across
+   * @param <E> the type of entities in the collection and underlying dataset
+   * @return an equivalent collection of entities partitioned for the view
+   * @see #partition(PCollection, View)
+   *
+   * @since 1.1.0
+   */
+  public static <E> PCollection<E> partition(PCollection<E> collection,
+                                             View<E> view,
+                                             int numWriters, int numPartitionWriters) {
+    //ensure the number of writers is honored whether it is per partition or total.
     DatasetDescriptor descriptor = view.getDataset().getDescriptor();
     if (descriptor.isPartitioned()) {
-      GetStorageKey<E> getKey = new GetStorageKey<E>(view);
-      PTable<GenericData.Record, E> table = collection
-          .by(getKey, Avros.generics(getKey.schema()));
-      PGroupedTable<GenericData.Record, E> grouped =
+      GetStorageKey<E> getKey = new GetStorageKey<E>(view,
+          numPartitionWriters > 0 ? numPartitionWriters : 1);
+      PTable<Pair<GenericData.Record, Integer>, E> table = collection
+          .by(getKey, Avros.pairs(Avros.generics(getKey.schema()), Avros.ints()));
+      PGroupedTable<Pair<GenericData.Record, Integer>, E> grouped =
           numWriters > 0 ? table.groupByKey(numWriters) : table.groupByKey();
       return grouped.ungroup().values();
     } else {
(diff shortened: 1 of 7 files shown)


kite-data/kite-data-crunch/src/main/java/org/kitesdk/data/crunch/CrunchDatasets.java
```diff
@@ -191,13 +191,43 @@ public static Target asTarget(URI uri) {
...
+   * @see #partition(PCollection, View)
+   *
+   * @since 1.1.0
+   */
+  public static <E> PCollection<E> partition(PCollection<E> collection,
+                                             View<E> view,
+                                             int numWriters, int numPartitionWriters) {
+
+    //ensure the number of writers is honored whether it is per partition or total.
+    int writers = Math.max(numPartitionWriters, numWriters);
```
rdblue: I think these should be separate configuration settings unless the dataset is unpartitioned. I'm fine taking the max of both for that case, but otherwise `writers` should be `numWriters`. So in the `isPartitioned` block, I think you should use `numWriters`, and use `writers` in the else block. Another alternative is to allow `numPartitionWriters` to override `numWriters` or just make keys with only the writer random number so the behavior is the same across both.

kite-data/kite-data-crunch/src/main/java/org/kitesdk/data/crunch/CrunchDatasets.java
```diff
@@ -271,11 +303,14 @@ public void initialize() {
               .getProvidedValues();
         }
       }
+      count = 0;
     }
 
     @Override
-    public AvroStorageKey map(E entity) {
-      return key.reuseFor(entity, provided, accessor);
+    public Pair<GenericData.Record, Integer> map(E entity) {
```
rdblue: Can Pair be serialized by Avro?

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/CompactCommand.java
```diff
@@ -43,6 +43,10 @@ public CompactCommand(Logger console) {
       description="The number of writer processes to use")
   int numWriters = -1;
 
+  @Parameter(names={"--num-partition-writers"},
```
rdblue: I'd like to make this a bit easier to understand. What about `--files-per-partition`?

rdblue:
Thanks @mkwhitacre! This looks pretty close. I would appreciate you adding those tests, if you don't mind. That was definitely an oversight.

kite-data/kite-data-crunch/src/main/java/org/kitesdk/data/crunch/CrunchDatasets.java
mkwhitacre: I think I get the behavior you are thinking but a few examples to confirm:

numWriters = 10
numPartitionWriters = 5
Partitioned: 10 reducers but partitions distribute across 5 reducers
Unpartitioned: 10 reducers

numWriters = 10
numPartitionWriters = 15
Partitioned: 10 reducers but partitions distribute across 15 values but will end up with at most 10 files.
Unpartitioned: 10 reducers

The second instance is why I picked max because then the files per partition would be honored.  

Or is your thought no-one would ever configure both num writers and num partition files?

kite-data/kite-data-crunch/src/main/java/org/kitesdk/data/crunch/CrunchDatasets.java
mkwhitacre: Yep, if it has a PType it can be serialized.

mkwhitacre:
If I promise to write tests but if don't get them done by 1.1.0 code freeze can this still make it in? :smile: 

I'll see if I can get some out tomorrow.

kite-data/kite-data-crunch/src/main/java/org/kitesdk/data/crunch/CrunchDatasets.java
rdblue: Sort of, but the values would be reversed. Right now, numWriters is the number of reducers. That should stay the same. Then numPartitionWriters (really should be filesPerPartition) would be the number of times a partition key is duplicated, ending up being the number of files. That means:

numWriters = 10, filesPerPartition=5 => 10 reducers, (5 * partition cardinality) unique reducer tasks

numWriters = 10, filesPerPartition=15 => 10 reducers, (15 * partition cardinality) unique reducer tasks

For the unpartitioned case, we can either use the same technique to generate reducer tasks, in which case we would limit the number of reducers to no more than the number of tasks. Or, we can just use the number of reducer tasks as the number of reducers for simplicity.

kite-data/kite-data-crunch/src/main/java/org/kitesdk/data/crunch/CrunchDatasets.java
rdblue: Awesome. How does that work? Not through reflection, I hope?

kite-data/kite-data-crunch/src/main/java/org/kitesdk/data/crunch/CrunchDatasets.java
mkwhitacre: But wouldn't in this situation:

numWriters = 10, filesPerPartition=15 => 10 reducers, (15 * partition cardinality) unique reducer tasks

It does not matter because since there are only 10 reducers the data will be at best distributed across 10 files?

kite-data/kite-data-crunch/src/main/java/org/kitesdk/data/crunch/CrunchDatasets.java
mkwhitacre: Not completely familiar with it but relevant code is here:
https://github.com/apache/crunch/blob/master/crunch-core/src/main/java/org/apache/crunch/types/avro/Avros.java#L490

kite-data/kite-data-crunch/src/main/java/org/kitesdk/data/crunch/CrunchDatasets.java
rdblue: Not, not really. Those 10 reducers will run through multiple tasks. With 15 tasks per partition, each one will get at least one task. Each task will create a writer, write the data, and close that writer, creating one file in one partition (because we know all the data is for the same partition).

mkwhitacre:
Ok added some tests.  I'm not completely sold on them b/c I was getting some funky results with regard to the number of files being produced.  I think that is just test setup (not enough data to properly distributed across the hashes or number of writers).  

I might not be able to get back to this till the end of the week unfortunately but wanted to put this out in case someone else has the cycles.

kite-data/kite-data-crunch/src/main/java/org/kitesdk/data/crunch/CrunchDatasets.java
rdblue: I didn't see that this uses Crunch's Pair class, rather than Kite's. That's why I was surprised it worked at first. :)

rdblue:
Thanks @mkwhitacre! This looks great. I'm working on making the tests work and I've already found a bug exposed by them.

rdblue:
Merged as https://github.com/kite-sdk/kite/commit/f8456578

Thanks @mkwhitacre!

PR #379, opened 2015-06-11
CDK-1016: Fix OutputFormat writing directly to datasets.

rdblue:
This happens only when a dataset instance is passed to the configuration
methods. The fix is to verify that the target dataset is not the root by
inspecting whether the partition key has any values.


rdblue:
@joey or @tomwhite could you review this for the 1.1.0 release? Thanks!

tomwhite:
Where did you see this happening? It looks fine to me, but probably could do with a test.

rdblue:
This was causing replace to fail because the tasks were writing directly to the dataset instead of the job dataset that replaces partitions.

rdblue:
Thanks Tom! I've merged this for now. #381 has tests that catch this case, which is how I found it. I'll merge those shortly.

PR #380, opened 2015-06-11
CDK-1017: Remove all target partitions in replace.

rdblue:
This removes any partitions in the view being replaced, even if the
replacement doesn't have the partition.


rdblue:
@tomwhite and @joey could you take a look at this for 1.1.0? Thanks!

tomwhite:
+1 looks good to me (once the tests are passing).

PR #382, opened 2015-06-11
CDK-1018: Avoid unnecessary copy in MR output format.

rdblue:
It appears that this was working around PARQUET-62, which fixed
dictionary support when incoming records are reused. Updating to 1.6.0
brings in the Parquet fix.

This also adds a property, kite.copyOutputRecords, that allows users to
control whether records should be copied. This defaults to false, but is
a good safety valve in case of other bugs like PARQUET-62.


joey:
+1 after Travis comes back green

rdblue:
Thanks, Joey!

PR #383, opened 2015-06-11
CDK-1019: Move Hive token setup into TransformTask.

rdblue:
This takes the delegation token code from the CopyCommand and moves it
into TransformTask so that all of the transform-based commands will work
with Hive.


joey:
+1

PR #386, opened 2015-06-18
CDK-991: Add size-based and time-based file rolling.

rdblue:
File size-based rolling works and is passing a new test for Avro, but is
disabled for Parquet because the appender has no reliable size estimate
for Parquet.

Time-based rolling uses a new SPI interface, ClockReady, which exposes a
method for passing time signals to implementing classes. This removes
the need for Kite to provide a thread-based check.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/ClockReady.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/ClockReady.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2015 Cloudera Inc.
+ *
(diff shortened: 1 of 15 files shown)


mkwhitacre:
Changes look good.  +1

rdblue:
Thanks @mkwhitacre!

I saw your question about when something wouldn't be ClockReady, which I think has been removed... but just in case the idea was that not all of the writers will be time-based. HBase, for example, doesn't need to be right now.

mkwhitacre:
yeah some of it seemed to be answered with the second commit (which I hadn't gotten to yet) :smile:

PR #390, opened 2015-06-29
KITE-1025 Use CombineFileInputFormat

gabrielreid:
Wrap Avro and Parquet inputs with CombineFileInputFormat so that
multiple small files don't result in multiple input splits and
MapReduce tasks.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/AbstractCombineFileRecordReader.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/AbstractCombineFileRecordReader.java
@@ -1,0 +1,86 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kitesdk.data.spi.filesystem;
+
+import java.io.IOException;
+import org.apache.hadoop.mapreduce.InputSplit;
+import org.apache.hadoop.mapreduce.RecordReader;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.apache.hadoop.mapreduce.lib.input.CombineFileSplit;
+import org.apache.hadoop.mapreduce.lib.input.FileInputFormat;
+import org.apache.hadoop.mapreduce.lib.input.FileSplit;
+
+/**
+ * Base class for wrapping file-based record readers with CombineFileInputFormat functionality. This allows multiple
+ * files to be combined into a single InputSplit, with the main pre-requisite being that the InputFormat being wrapped
+ * must make use of FileInputSplits.
+ * <p/>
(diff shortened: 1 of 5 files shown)


rdblue:
@gabrielreid, thanks for working on this. It will be great to have combine support.

Can you write a little on how it is intended to work? I noticed that this removes the check in Crunch. What changes for regular MR and for Crunch?

gabrielreid:
Sure, sorry for not adding some additional explanation previously.

The way that Crunch works (in recent versions) is that it combines multiple InputSplits into one (via CrunchCombineFileInputFormat) if both of the following are true:
* The underlying InputFormat is a subclass of FileInputFormat
* RuntimeParameters.DISABLE_COMBINE_FILE is false

Before this commit, Kite correctly set DISABLE_COMBINE_FILE to false for Parquet and Avro files, but due to the underlying input format not being a subclass of FileInputFormat anymore, Crunch didn't wrap it in CrunchCombineFileInputFormat, so no combined splits were created.

This commit introduces similar CombineFileInputFormat logic into Kite, as letting Crunch handle the combining is no longer feasible.

CombineFileInputFormat is used for Avro and Parquet files. This involves two things:
* overriding getSplits for Avro and Parquet to pass it off to CombineFileInputFormat
* wrapping the Avro and Parquet RecordReaders with implementations that can handle the CombineSplits from CombineFileInputFormat

The JSON and CSV formats aren't changed, because it appears that their record readers are instantiated with a path to read from, which wouldn't directly map to the way CombineFileInputFormat works (as it relies on the initialize() method in a RecordReader to set up the path to read from). With some more work I think it would be feasible to add combine support for JSON and CSV however.

The underlying Combine functionality is automatically used in both Crunch and normal MR jobs.

Note that I've just updated this commit, as I discovered that there was an issue with it when multiple sources were being read in a single Crunch task. CrunchRecordReader has some logic that specifically checks if an incoming InputSplit is a CombineFileSplit. Seeing as Kite is now generating CombineFileSplits, Crunch was incorrectly interpreting these input splits as something created by Crunch. I've worked around this by adding AbstractKiteCombineFileInputFormat and KiteCombineFileSplit, which only exist to make Crunch treat them as "normal" input splits. It's not nice to add all that stuff in there, but I don't see any other way to get around it -- probably wouldn't be bad to update Crunch to do the same thing (i.e. have its own version of CombineFileSplit), and then this workaround could be removed from Kite.

gabrielreid:
Weird, the Travis build is failing now with a timeout on org.kitesdk.data.crunch.TestCrunchDatasetsHive. Building everything under kite-data with the cdh4 profile seems to work fine for me -- is that a test that is a little unreliable?

rdblue:
Yeah, we've seen flakiness with Travis on that test before. I restarted it and that usually fixes the problem.

mkwhitacre:
Is there something outstanding on this issue or can it possibly be merged into master?  It would definitely help with some situations we are seeing with hourly partitions and improve how fast we can roll them up into daily partitions.

rdblue:
@mkwhitacre, I just need to review this and have been busy lately. I'll hopefully get to reviews this week.

mkwhitacre:
Cool, was looking at improving the performance of some stuff and remembered @gabrielreid did this PR.

mkwhitacre:
+1 to this change.  Going to try building with these changes to verify it improve performance due to lots of mappers.

mkwhitacre:
Saw a noticeable diff when running with this code.  Specifically in our dev environment where we have lots of small files because of hourly extracts and low traffic volume of inflowing data (100k vs 1k of mappers).

gabrielreid:
Thanks for taking it for a spin @mkwhitacre. That's pretty much in line with what I've seen running it as well.

rdblue:
Merged as https://github.com/kite-sdk/kite/commit/e92bf712. Thanks @gabrielreid, this is a really valuable new feature! And thanks @mkwhitacre for reviewing and testing it out!

PR #393, opened 2015-07-07
Add partitionAndSort

tomwhite:


rdblue:
Looks like tests are failing because of findbugs errors.

rdblue:
Could you open an issue for this?

kite-data/kite-data-crunch/src/main/java/org/kitesdk/data/crunch/CrunchDatasets.java
```diff
@@ -231,6 +232,38 @@ public static Target asTarget(URI uri) {
...
+  }
+
+  public static <E, K> PCollection<E> partitionAndSort(PCollection<E> collection,
+      View<E> view, MapFn<E, GenericData.Record> getSortKeyFn, Schema sortKeySchema,
+      int numWriters) {
+    return partitionAndSort(collection, view, getSortKeyFn, sortKeySchema, numWriters, 1);
+  }
+
+  public static <E, K> PCollection<E> partitionAndSort(PCollection<E> collection,
+      View<E> view, MapFn<E, GenericData.Record> getSortKeyFn, Schema sortKeySchema,
```
rdblue: Can this be something other than GenericData.Record? It would be nice to parameterize this and allow the caller to pass in a PType<S> for the sort key rather than a Schema. There are a lot of cases where the caller just wants to extract a single field for the sort.

Speaking of extracting a single field, that may be a valuable variant as well: `partitionAndSort(PCollection, View, String fieldName)`  That doesn't need to go into this commit though.

tomwhite: I agree it would be better to make this more flexible. I'll have a look at doing that.

mkwhitacre:
@tomwhite is this PR still alive?

tomwhite:
@mkwhitacre I haven't had a chance to update it with Ryan's feedback. I probably won't for a bit, so feel free to take it over if you need it.

PR #396, opened 2015-07-14
Kite-1042: OOZIE_ACTION_CONF_XML added to kite-dataset along with 'config' option

mladkov:
When executing the `kite-dataset` cli command through Oozie, we need to pass into the `hadoop jar` command the `-conf` parameter to the location of `OOZIE_ACTION_CONF_XML`, which is set by Oozie.

Additionally, we provide the ability for the user to specify a `config` option like so:

 `config="-D key=value" ./kite-dataset ...`

And those parameters will be passed into the `hadoop jar` command.  This is similar to the `flags` option and the `debug` option that the utility supplies.

### DIFF
--- a/kite-tools-parent/kite-tools/src/main/sh/dataset.sh
+++ b/kite-tools-parent/kite-tools/src/main/sh/dataset.sh
@@ -119,6 +119,17 @@ if [ ! -z "$KITE_USER_CLASSPATH" ]; then
 fi
 debug "Using HADOOP_CLASSPATH=${HADOOP_CLASSPATH}"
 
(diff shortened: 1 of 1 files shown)


mladkov:
I see that my build had failed - but that really only happened it seems due to environment errors on the build machine. What do we do next? How can I re-submit the build? Do I have to generate a new pull request and close this one?

mladkov:
I see the build has now passed - thanks for restarting the build.

Is there anything else needed here, or will this request actually get pulled to the Kite project?

rdblue:
Merged as https://github.com/kite-sdk/kite/commit/38c25cc1. Thanks @mladkov!

PR #398, opened 2015-07-15
Kite-1046: Add num-records option to json-schema command

mladkov:
This code piggy backs on the existing strategy for parsing new options.

Only difference here is adding the "arity" parameter, to force users to enter exactly 1 value.

Here's how the help text now looks:

```
Usage: ./kite-dataset [general options] json-schema <sample json path> [command options]

  Description:

    Build a schema from a JSON data sample

  Command options:

    -n, --num-records
	Number of records to derive schema from (default: 10)
    -o, --output
	Save schema avsc to path
  * --class, --record-name
	A name or class for the result schema
    --minimize
	Minimize schema file size by eliminating white space

  * = required
```

If you don't enter a value in, you'll see this:

```
Expected a value after parameter --num-records
```

And if you specify something other than a valid integer is gives you this:

```
"-n": couldn't convert "blah" to an integer
```

### DIFF
--- a/kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/JSONSchemaCommand.java
+++ b/kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/JSONSchemaCommand.java
@@ -65,6 +65,15 @@ public class JSONSchemaCommand extends BaseCommand {
       description="Minimize schema file size by eliminating white space")
   boolean minimize=false;
 
+  @edu.umd.cs.findbugs.annotations.SuppressWarnings(
+      value="UWF_NULL_FIELD",
+      justification = "Field set by JCommander")
+  @Parameter(names={"-n", "--num-records"},
+             description="Number of records to derive schema from",
+             arity=1)
+  int numRecords = 10;  // By default we look at first 10 records
+
+
   @Override
   public int run() throws IOException {
     Preconditions.checkArgument(samplePaths != null && !samplePaths.isEmpty(),
@@ -74,7 +83,7 @@ public class JSONSchemaCommand extends BaseCommand {
 
     // assume fields are nullable by default, users can easily change this
     Schema sampleSchema = JsonUtil.inferSchema(
-        open(samplePaths.get(0)), recordName, 10);
+        open(samplePaths.get(0)), recordName, numRecords);
 
     if (sampleSchema != null) {
       output(sampleSchema.toString(!minimize), console, outputPath);


kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/JSONSchemaCommand.java
```diff
@@ -65,6 +65,15 @@ public JSONSchemaCommand(Logger console) {
       description="Minimize schema file size by eliminating white space")
   boolean minimize=false;
 
+  @edu.umd.cs.findbugs.annotations.SuppressWarnings(
+      value="UWF_NULL_FIELD",
+      justification = "Field set by JCommander")
+  @Parameter(names={"-n", "--num-records"},
+             description="Number of records to derive schema from",
+             arity=1)
+  Integer numRecords = 10;  // By default we look at first 10 records
```
rdblue: Why use `Integer` here and `intValue` later?

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/JSONSchemaCommand.java
```diff
@@ -65,6 +65,15 @@ public JSONSchemaCommand(Logger console) {
       description="Minimize schema file size by eliminating white space")
   boolean minimize=false;
 
+  @edu.umd.cs.findbugs.annotations.SuppressWarnings(
+      value="UWF_NULL_FIELD",
+      justification = "Field set by JCommander")
+  @Parameter(names={"-n", "--num-records"},
+             description="Number of records to derive schema from",
+             arity=1)
```
rdblue: Isn't `arity=1` the default?

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/JSONSchemaCommand.java
mladkov: Just because I am new to JCommander, and I think their example when it came to integers was using the Object Integer instead of the primitive type.  I changed it to int, and validated it works fine. So I'll checkin that adjustment.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/JSONSchemaCommand.java
mladkov: Seems like it is. However, what's different is when you do NOT explicitly specify it, the help output looks like this:
```
-n, --num-records
	Number of records to derive schema from
```
Compared to when you explicitly state it, JCommander seems to be able to pull out the value you initialized the variable with, and show the user what the default is:

```
-n, --num-records
	Number of records to derive schema from (default: 10)
```
I thought that was neat. Are you ok with showing this default?

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/JSONSchemaCommand.java
rdblue: Thanks! You can force-push to update the PR.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/JSONSchemaCommand.java
rdblue: Yeah, I like the default. I think that might not be JCommander though, I think it is my help code: https://github.com/kite-sdk/kite/blob/master/kite-minicluster/src/main/java/org/kitesdk/minicluster/cli/Help.java#L123

We should probably fix that eventually, but arity=1 is a good work-around for now.

mladkov:
Ryan, what happens next? Will these changes be pulled or rejected?  When will the pull actually happen if approved?

rdblue:
Merged as https://github.com/kite-sdk/kite/commit/ff6afe42. Thanks @mladkov!

PR #399, opened 2015-07-16
KITE-1035 read hive config from HCatAccessorService in KiteURIHandler

bbrownz:
https://issues.cloudera.org/browse/KITE-1035

Read the hive configuration from the HCatAccessorService in the KiteURIHandler to handle cases like view URIs without the metastore URI built in or more complex situations where kerberos needs to be configured for Hive.

Since this makes use of the HCatAccessorService configuration it's limited to hive configuration. A potential alternative can be seen in the commit [here](https://github.com/bbrownz/kite/commit/5d5e16913) where a configuration service is added specifically for Kite. The configuration is read from a list of potential configurations. This has the additional downside of requiring to consumer to configure the location of these config files as well as specify that Oozie should load the service.

### DIFF
--- a/kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
+++ b/kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
@@ -20,6 +20,7 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.oozie.ErrorCode;
@@ -27,7 +28,9 @@ import org.apache.oozie.XException;
 import org.apache.oozie.action.hadoop.LauncherURIHandler;
 import org.apache.oozie.dependency.URIHandler;
 import org.apache.oozie.dependency.URIHandlerException;
+import org.apache.oozie.service.HCatAccessorService;
 import org.apache.oozie.service.HadoopAccessorException;
+import org.apache.oozie.service.Services;
 import org.apache.oozie.service.URIHandlerService;
 import org.apache.oozie.util.XLog;
 import org.kitesdk.data.DatasetException;
@@ -36,6 +39,7 @@ import org.kitesdk.data.Datasets;
 import org.kitesdk.data.Signalable;
 import org.kitesdk.data.URIBuilder;
 import org.kitesdk.data.View;
+import org.kitesdk.data.spi.DefaultConfiguration;
 
 /**
  * A Kite URI handler that works with {@link Signalable} views.
@@ -49,6 +53,8 @@ public class KiteURIHandler implements URIHandler {
 
   private Set<String> supportedSchemes;
   private List<Class<?>> classesToShip;
+
+  private AtomicBoolean confLoaded = new AtomicBoolean(false);
 
   @Override
   public void init(final Configuration conf) {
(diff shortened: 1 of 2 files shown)


kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
```diff
@@ -101,9 +106,38 @@ public Context getContext(final URI uri, final Configuration conf,
...
+    confLoaded = true;
+  }
+
   @SuppressWarnings("rawtypes")
   @Override
   public boolean exists(final URI uri, final Context context) throws URIHandlerException {
+    // load the configuration if it hasn't been yet
+    if(!confLoaded) {
+      loadConfiguration();
+    }
```
bbrownz: It's worth mentioning that I load the configuration here in the exists method instead of the init method. In my testing I found that Oozie would initialize the URIHandlers before it initialized the services and the desired services would not be available.

kite-data/kite-data-oozie/src/main/java/org/kitesdk/data/oozie/KiteURIHandler.java
```diff
@@ -101,9 +107,35 @@ public Context getContext(final URI uri, final Configuration conf,
...
+  }
+
   @SuppressWarnings("rawtypes")
   @Override
   public boolean exists(final URI uri, final Context context) throws URIHandlerException {
+    // load the configuration if it hasn't been yet
+    if(confLoaded.compareAndSet(false, true)) {
+      loadConfiguration();
+    }
```
bbrownz: It's worth mentioning that I load the configuration here in the exists method instead of the init method. In my testing I found that Oozie would initialize the URIHandlers before it initialized the services and the desired services would not be available.

rdblue:
Looks okay to me. @rkanter, could you take a look?

kite-data/kite-data-oozie/src/test/java/org/kitesdk/data/oozie/TestKiteURIHandler.java
```diff
@@ -215,6 +238,90 @@ public void existsForNonReadiableView() throws URIHandlerException, URISyntaxExc
...
+    conf.writeXml(new FileOutputStream(hadoopConfTarget));
+
+    Configuration hiveConf = new Configuration(false);
+    hiveConf.set("test.property", "test.value");
+    hiveConf.writeXml(new FileOutputStream(hiveConfTarget));
+
+    // set to the temp directory
+    System.setProperty("oozie.home.dir", serviceTempDir.getAbsolutePath());
+
+    Services services = new Services();
```
rkanter: We need to make sure to destroy the Services object after each test.  We used to have a lot of flakey tests in Oozie because we were missing some of these, so Services would bleed into other tests causing all kinds of weird and hard-to-diagnose problems.

We typically take care of this by creating it in the setUp() method and destroying in the tearDown() method.  
```
private Services services;
@Override
protected void setUp() throws Exception {
   super.setUp();
    services = new Services();
    services.init();
}
@Override
protected void tearDown() throws Exception {
    services.destroy();
    super.tearDown();
}
```
For tests that require more flexibility or creating/destroying the Services, we usually use a try-finally block:
```
Services services = null;
try {
   services = new Services();
   services.init();
   // do stuff
} finally {
   if (services != null) {
      services.destroy();
   }
}
```

rkanter:
It might make sense to go with the KiteConfigurationService in the other PR.  It will allow Kite to be more flexible in the future.  For example, if you want to use more than just Hive datasets, you may need other config files, which we wouldn't want the HCatAccessorService to load; with your own KiteConfigurationService, that wouldn't be a problem.  And switching from HCatAccessorService to KiteConfigurationService later could be considered backwards incompatible because the oozie-site config property would change.

kite-data/kite-data-oozie/src/test/java/org/kitesdk/data/oozie/TestKiteURIHandler.java
bbrownz: I took the route up on [line 113](https://github.com/kite-sdk/kite/pull/399/files#diff-bc98164dc4c52f3fd75c88e983b578eaR113) of calling the static "get" on the Services object to see if it had been initialized. On init it looks like the Services object just sets itself for singleton access. I then call destroy on the service object.

This is probably a bit harder to trace, it requires that you're familiar enough with the services object to catch that it's destroying the same object. It might be better for that reason just to keep a reference.

rkanter: Oh, that's fine.  I missed the @After there.  You can ignore my comment then.

rdblue:
It sounds like we should take @rkanter's suggestion to add a KiteConfigurationService. Any thoughts @bbrownz?

bbrownz:
Agreed, I've been meaning to get around to polishing up that branch so we can close this PR out in favor of the other one. Hopefully tomorrow :)

bbrownz:
PR opened for the alternative implementation #405

PR #406, opened 2015-07-30
KITE-1013: Configurable Sample-Based Type Inferencing for JSON and CSV

bdrillard:
These commits provide a patch for KITE-1013, to add configurable, sample-based type inferencing for JSON and CSV. It adds the command-line flags -n, --samples which lets the user specify the number of rows or objects to sample when inferencing types for schemas. 

KITE-1013: adding command-line param to csv-schema
KITE-1013: requiring at least one sample row
KITE-1013: adding samples option to JSON-schema
KITE-1013: testing CSVSchema on too few samples
KITE-1013: removing debug log, fixing test typo
KITE-1013: squashing commit history

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/CSVUtil.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/CSVUtil.java
@@ -165,7 +165,7 @@ public class CSVUtil {
             if (types[i] != null) {
               values[i] = line[i];
             }
-          } else if (mergeType(prevType, currType)) {
+          } else if (mergeTypes(prevType, currType)) {
             types[i] = currType;
             values[i] = line[i];
           }


rdblue:
Thanks for taking the time to contribute, @bdrillard!

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/CSVProperties.java
```diff
@@ -35,6 +35,7 @@
   public static final String HEADER_PROPERTY = "kite.csv.header";
   public static final String HAS_HEADER_PROPERTY = "kite.csv.has-header";
   public static final String LINES_TO_SKIP_PROPERTY = "kite.csv.lines-to-skip";
+  public static final String LINES_TO_SAMPLE_PROPERTY = "kite.csv.lines-to-sample";
```
rdblue: I don't think this is something that needs to be in the CSVProperties. Those are for describing how you would the contents of a CSV file and sampling isn't something you need to know to decode the CSV. Just passing the number of lines to use like we do in [JsonUtil](https://github.com/kite-sdk/kite/blob/master/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/JsonUtil.java#L517) should be sufficient.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/CSVUtil.java
```diff
@@ -133,19 +134,23 @@ private static Schema inferSchemaInternal(String name, InputStream incoming,
...
-            types[i] = inferFieldType(line[i]);
+          Schema.Type prevType = types[i];
+          Schema.Type currType = inferFieldType(line[i]);
+          if (prevType == null) {
+            types[i] = currType;
             if (types[i] != null) {
-              // keep track of the value used
               values[i] = line[i];
             }
+          } else if (promoteType(prevType, currType)) {
```
rdblue: I think this would be better if you changed `promoteType` to `mergeTypes` (I think there might already be one somewhere. Maybe SchemaUtil or JsonUtil).

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestCSVSchemaInference.java
```diff
@@ -284,6 +296,32 @@ public void testSchemaInferenceSkipHeader() throws Exception {
   }
 
   @Test
+  public void testSchemaTypePromotion() throws Exception {
+        LOG.debug("########## testSchemaTypePromotion ##############");
```
rdblue: No need for the debug lines here, please.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/CSVSchemaCommand.java
```diff
@@ -79,6 +79,9 @@ public CSVSchemaCommand(Logger console) {
   @Parameter(names="--skip-lines", description="Lines to skip before CSV start")
   int linesToSkip = 0;
 
+  @Parameter(names={"-n", "--samples"}, description="Number of lines to sample when inferencing schema")
```
rdblue: We just added --num-records to the JSON equivalent. Could you copy that so the doc and options match?

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/CSVSchemaCommand.java
```diff
@@ -109,6 +114,7 @@ public int run() throws IOException {
         .header(header)
         .hasHeader(!noHeader)
         .linesToSkip(linesToSkip)
+        .linesToSample(linesToSample)
```
rdblue: This should be passed via the `inferNullableSchema` call below.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/JSONSchemaCommand.java
```diff
@@ -65,16 +65,22 @@ public JSONSchemaCommand(Logger console) {
...
         "Sample JSON path is required");
     Preconditions.checkArgument(samplePaths.size() == 1,
         "Only one JSON sample can be given");
+    Preconditions.checkArgument(objectsToSample > 0,
+        "At least one object must be sampled");
 
     // assume fields are nullable by default, users can easily change this
     Schema sampleSchema = JsonUtil.inferSchema(
-        open(samplePaths.get(0)), recordName, 10);
+        open(samplePaths.get(0)), recordName, objectsToSample);
```
rdblue: This just got merged from another PR, so no need to do it anymore. Thanks for fixing it though!

rdblue:
Nice work! Just a few things to fix and we'll get this merged. Thanks!

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/CSVUtil.java
bdrillard: I've changed the name as you suggested. The code I've used in the CSVUtil mergeTypes method is similar to the same method in the SchemaUtil file, but I've kept the one I've written because it handles merging of previous numeric types to strings when necessary. I do so because CSV files can be notorious for having odd string flags in otherwise numeric columns, so I try to allow for strings as a 'most generic' type when necessary. Let me know if you disagree with this reasoning, I could rearrange it to make use of the other mergeTypes method if its is preferred.

PR #407, opened 2015-08-26
KITE-1040: Cache Hive MetaStore connections.

rdblue:
This adds a MetaStoreUtil.get method that keeps track of util objects
(and the internal MetaStore connection) by connection URIs. These are
safe to share because they are thread-safe.


rdblue:
@tomwhite could you take a look at this? This is to avoid reconnecting to the metastore all the time in Flume and other clients that may refresh dataset references from time to time.

tomwhite:
+1 LGTM

PR #408, opened 2015-08-26
KITE-1065: Add support for Oozie delegation tokens.

rdblue:
Oozie passes delegation tokens via HADOOP_TOKEN_FILE_LOCATION. This adds
support to pick up those tokens and add them to the DefaultConfiguration.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DefaultConfiguration.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DefaultConfiguration.java
@@ -38,16 +38,22 @@ public class DefaultConfiguration {
   private static Configuration conf;
 
   static{
-    //read system property for Oozie Action Configuration XML
+    // read system property for Oozie Action Configuration XML
     String oozieActionXml = System.getProperty("oozie.action.conf.xml");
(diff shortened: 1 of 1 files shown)


rdblue:
@rkanter and @bbrownz, can you review this?

rkanter:
Looks good to me +1

bbrownz:
+1 from me as well. Tagging @mkwhitacre as well. I believe he made the addition for the static block on https://issues.cloudera.org/browse/KITE-462 although I'm not sure if he has run into issues with the token file as well.

bbrownz:
Site note: it looks like KITE-462 might have missed being tagged with 1.1.0 where Micah's changes went in.

mkwhitacre:
+1 for the change as I'm pretty sure it is innocuous or should have minimal impact.  I'm curious where this error presented itself because we've been running without it on a secure cluster and communicating between two instances just fine. 
 
Outside of Kite the only place I know this is a requirement is HBase[1]
[1] - https://github.com/apache/hbase/blob/4f410e6bd5406eacbfae98d21786dfc24cb57f1b/hbase-server/src/main/java/org/apache/hadoop/hbase/mapred/TableMapReduceUtil.java#L265

PR #410, opened 2015-09-10
KITE-1068: SolrCellMorphlineTests fails on Locales with non-Arabic di…

grchanan:
SolrCellMorphlineTest (and I assume other tests that derive from AbstractSolrMorphlineTest, though I didn't explicitly test them) fail on Locales that use non-Arabic numbers.


rdblue:
Committed as https://github.com/kite-sdk/kite/commit/ae63d6b942e1f75600c60218687f6f505971ca35

PR #411, opened 2015-09-15
KITE-1071: Use exit 1 when Hadoop isn't found by CLI.

rdblue:


tomwhite:
+1 when tests come back green.

PR #413, opened 2015-09-16
KITE-1057: Fix signal path for single data file.

rdblue:
The CLI will create a dataset around a single data file, but the
FileSystemDataset code assumed that the path it uses is a directory.
This caused errors because the signal manager attempted to use the
incoming data file path as a directory that potentially contained the
signals folder. This updates the FileSystemDataset to use the data
file's parent for the signal path.


tomwhite:
Looks good. What testing have you done?

rdblue:
Tested on 5.5 and 5.4.6. That's how I found the other issue. :)

tomwhite:
+1

PR #414, opened 2015-09-16
KITE-1073: Remove work-around setting default FS in CLI.

rdblue:
This removes the work-around added for KITE-898, where job submission
was failing when importing local-to-HDFS unless the default FS was
local. This is no longer needed and is causing a different failure.


tomwhite:
+1

jastang:
@rdblue @tomwhite is this related to the job commit error from csv-import as well? If so, do we need 1.2.0? This is when i'm importing a CSV residing in HDFS:
 
org.apache.hadoop.mapreduce.v2.app.commit.CommitterEventHandler: Could not commit job
java.lang.IllegalArgumentException: Wrong FS: file:/tmp/ingestion/.temp/job_1481125587968_0017/mr/job_1481125587968_0017/4d9cf95f-fe9c-48d1-a619-1750e96f2f1f.avro, expected: hdfs://sandbox.hortonworks.com:8020

PR #415, opened 2015-09-17
KITE-1076: Fix flaky Crunch-Hive test.

rdblue:
The test fails in job setup using the local job runner when the HBase
test has been run first. It appears to be a bug in the test runner code
that can be avoided by not reusing test VMs for the Crunch module.


rdblue:
@tomwhite do you think this is a reasonable fix?

tomwhite:
+1

PR #416, opened 2015-09-23
KITE-1079: Improve URI debugging with error messages.

rdblue:
This adds a map to DynConstructor.Builder that keeps all of the reasons
why classes weren't loaded. In the case where one is present but can't
be loaded because a dependency is missing, it is obvious what needs to
be added to the classpath.


tomwhite:
+1

PR #417, opened 2015-09-23
KITE-1080: Update to org.apache Parquet 1.8.0.

rdblue:


kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/MarkerRange.java
```diff
@@ -201,9 +201,9 @@ private Boundary() {
     }
 
     public Boundary(MarkerComparator comparator, Marker bound, boolean isInclusive) {
-      parquet.Preconditions.checkArgument(comparator != null,
+      Preconditions.checkArgument(comparator != null,
```
tomwhite: What's the difference between the two?

tomwhite:
One small question, but looks fine otherwise.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/MarkerRange.java
rdblue: Parquet doesn't want to depend on guava, so it defined its own precondition helpers. It was an accident to use the Parquet one here (or anywhere outside of Parquet).

tomwhite:
I see it now (I had it the wrong way round in my head).

+1

rdblue:
Merged. Thanks, Tom!

PR #419, opened 2015-09-25
KITE-1083: Add single jars to DistCache for Hive.

rdblue:
This changes the way jobs submitted by the CLI are configured.
Previously, the entire lib directory for Hive was added to the
distributed cache. This caused long job and task startup times and
exposed some conflicting jar problems. This commit updates the setup so
that individual jars are added for classes needed for interacting with
the Hive MetaStore. In cases where the job is local or the job isn't
interacting with Hive, this doesn't add Hive dependencies to the
distributed cache at all.

### DIFF
--- a/kite-tools-parent/kite-tools/src/main/java/org/kitesdk/tools/TransformTask.java
+++ b/kite-tools-parent/kite-tools/src/main/java/org/kitesdk/tools/TransformTask.java
@@ -21,6 +21,9 @@ import com.google.common.collect.Iterables;
 import java.io.IOException;
 import java.net.URI;
 import org.apache.avro.generic.GenericRecord;
+import org.apache.commons.codec.Encoder;
+import org.apache.commons.compress.compressors.CompressorInputStream;
+import org.apache.commons.logging.Log;
 import org.apache.crunch.DoFn;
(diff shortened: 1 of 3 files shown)


rdblue:
Notes:
* This updates the default Hive dependency to 1.1.0
* There is added handling for thrift that is required for CDH 5.4+ clusters
* This was tested on CDH 5.4

tomwhite:
This seems very brittle. If a dependency changes then it won't be obvious that TransformTask needs to be updated.

rdblue:
I agree that this is brittle, but we need to connect to the metastore in setup and commit tasks. Adding the Hive lib directory was worse because clusters that used parcels ended up including the entire jars directory of the parcel. That ended up causing version conflicts and causing the jobs to take forever because so many jars were added in the distributed cache.

We can try to come up with a better solution, but for now I know this works and is better than the old method. We should be able to catch dependency changes with integration testing.

tomwhite:
I haven't got a better suggestion, so I agree that it's OK to go in. Could you add a comment to the POM dependencies that points to this class so anyone changing it know to look there too.

rdblue:
Good idea, I will.

PR #420, opened 2015-09-29
KITE-1077: Added dry run option to the delete command

mkwhitacre:


bbrownz:
+1

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/DeleteCommand.java
```diff
@@ -31,6 +32,10 @@
   @Parameter(description = "<dataset or view>")
   List<String> targets;
 
+  @Parameter(names="--dryRun",
```
rdblue: The convention for other arguments is lower case separated by dash. Could you change this to `--dry-run`?

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/DeleteCommand.java
```diff
@@ -46,11 +51,24 @@ public int run() throws IOException {
...
+        if (!dryRun){
+          Datasets.delete(uriOrName);
+        }
       } else {
-        getDatasetRepository().delete(namespace, uriOrName);
+        if(!dryRun) {
+          getDatasetRepository().delete(namespace, uriOrName);
+        }
       }
       console.debug("Deleted {}", uriOrName);
```
rdblue: I think this should change depending on whether this is a dry run. If so, it should say "Would have deleted {}" or something similar to signal that no action was taken.

kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/DeleteCommand.java
```diff
@@ -46,11 +51,24 @@ public int run() throws IOException {
         View view = Datasets.load(uriOrName);
         Preconditions.checkArgument(viewMatches(view.getUri(), uriOrName),
             "Resolved view does not match requested view: " + view.getUri());
-        view.deleteAll();
+
+        if(console.isDebugEnabled()){
+          for(PartitionView<?> pv: (Iterable<PartitionView<?>>) view.getCoveringPartitions()){
+            console.debug("Deleting partition {}", pv.getLocation());
```
rdblue: This assumes that the underlying view is a FS view and that it is aligned with partition boundaries. If either of those isn't true then this will either throw a ClassCastException or lie about what would have been deleted. We could add a `canDelete()` method to the view if necessary or just special case partitions?

mkwhitacre: I guess I'm not following, are you saying getCoveringPartitions() doesn't return a partition boundary?  Is this because if someone gave a URI with a constraint you didn't partition on it would still be part of the URI?

rdblue: The partitions returned by `getCoveringPartitions` may contain values that aren't in the view. For example, if I have a view for all of today but I'm partitioning by months, I get a partition for all of October.

When deleting data, we check that the view we are deleting aligns with partitions: meaning that the view is exactly equal to the partitions we will delete, with no data in the partitions that doesn't match the view. Without doing that check here, we would report that more data would be deleted than actually is deleted.

Does that make more sense?

mkwhitacre: Last commit is adding the canDelete method.  Let me know if that is what you were thinking.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/AbstractDataset.java
```diff
@@ -150,6 +150,12 @@ public boolean deleteAll() {
   }
 
   @Override
+  public boolean canDelete() {
+    throw new UnsupportedOperationException(
+            "This Dataset does not support checking deletion");
```
rbrush: Should this just return false rather than throwing and exception? I don't see user code making a distinction between "can't delete this" and "delete operation not supported for this", but there may be a flow I'm missing.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/AbstractRefinableView.java
```diff
@@ -153,6 +153,12 @@ public boolean deleteAll() {
   }
 
   @Override
+  public boolean canDelete() {
+    throw new UnsupportedOperationException(
+            "This Dataset does not support checking bulk deletion");
+  }
```
rbrush: Same question as in AbstractDataset: should this just return false?

rbrush:
Posted a couple questions above but otherwise this looks good to me.

PR #421, opened 2015-10-08
KITE-1081 FS writers to use the current view schema

megnataraj:


kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemView.java
```diff
@@ -118,7 +118,7 @@ private FileSystemView(FileSystemView<?> view, Schema schema, Class<E> type) {
       writer = PartitionedDatasetWriter.newWriter(this);
     } else {
       writer = FileSystemWriter.newWriter(
-          fs, root, -1, -1 /* get from descriptor */, dataset.getDescriptor());
+          fs, root, -1, -1 /* get from descriptor */, dataset.getDescriptor(), this.getSchema());
```
rdblue: The `AbstractRefinableView#getSchema` method returns the accessor's read schema. That's useful for callers, but the writers should be initialized using `getAccessor().getWriteSchema()`. The two schemas can differ slightly for reflected records.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/PartitionedDatasetWriter.java
```diff
@@ -267,7 +267,7 @@ public DatasetWriterCacheLoader(FileSystemView<E> view, ConfAccessor conf) {
           dataset.getFileSystem(),
           new Path(dataset.getDirectory(), partition),
           conf.getRollIntervalMillis(), conf.getTargetFileSize(),
-          dataset.getDescriptor());
+          dataset.getDescriptor(), view.getSchema());
```
rdblue: These should also use the accessor's write schema.

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestAvroWriter.java
```diff
@@ -44,9 +44,9 @@
                 String.valueOf(32 * 1024 * 1024)) // 32 MB
             .schema(schema)
             .format("avro")
-            .build());
+            .build(), TEST_SCHEMA);
   }
-
+
```
rdblue: Nit: whitespace-only change.

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestFileSystemWriters.java
```diff
@@ -52,7 +56,7 @@
   protected FileSystem fs = null;
   protected Path testDirectory = null;
   protected FileSystemWriter<Record> fsWriter = null;
-
+
```
rdblue: Nit: whitespace-only change.

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestFileSystemWriters.java
```diff
@@ -45,6 +45,10 @@
       .requiredLong("id")
       .requiredString("message")
       .endRecord();
+  
+  public static final Schema SCHEMA = SchemaBuilder.record("Message").fields()
+	      .requiredLong("id")
```
rdblue: Nit: indentation is off. This should match the indentation of TEST_SCHEMA.

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestFileSystemWriters.java
```diff
@@ -95,6 +99,38 @@ public void testBasicWrite() throws IOException {
...
+    Assert.assertEquals("Should contain no visible files", 0, stats.length);
+    stats = fs.listStatus(testDirectory);
+    Assert.assertEquals("Should contain a hidden file", 1, stats.length);
+
+    fsWriter.close();
+
+    stats = fs.listStatus(testDirectory, PathFilters.notHidden());
+    Assert.assertEquals("Should contain a visible data file", 1, stats.length);
+
+    DatasetReader<Record> reader = newReader(stats[0].getPath(), TEST_SCHEMA);
```
rdblue: This should actually fail because SCHEMA isn't compatible with TEST_SCHEMA; the message field is required for the latter, but missing in the "old" schema. I think we need to understand why this schema isn't passed all the way through as the read schema.

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestAvroWriter.java
```diff
@@ -44,9 +44,9 @@
                 String.valueOf(32 * 1024 * 1024)) // 32 MB
             .schema(schema)
             .format("avro")
-            .build());
+            .build(), TEST_SCHEMA);
```
rdblue: This should be the schema passed into `newWriter`.

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestParquetWriter.java
```diff
@@ -40,7 +40,7 @@
                 "kite.writer.roll-interval-seconds", String.valueOf(1))
             .schema(schema)
             .format("parquet")
-            .build());
+            .build(), TEST_SCHEMA);
```
rdblue: This should also be the schema passed in.

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestAvroWriter.java
megnataraj: Here, I am not passing the same schema as the dataset one so that I can test the scenario where the dataset is updated with the new schema but data is written with the old schema

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestFileSystemWriters.java
megnataraj: Yeah, you're right in reality it should fail the compatibility check but this test case isn't failing as it is a unit test case and the compatibility check happens much before the control goes to FileSystemWriter. Granted I got it wrong here, instead I should have updated the dataset schema and write with the old schema. Coming to the DatasetReader, I think the writer schema is passed on to the reader schema[1]. However, more clarification on this will help me better understand it.  

[1]: https://github.com/kite-sdk/kite/blob/master/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/DataModelUtil.java#L129

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestAvroWriter.java
rdblue: This helper method shouldn't directly access TEST_SCHEMA. If you want to be able to use it to test this case, then lets use separate datasetSchema and writerSchema. Also, I think you have it backwards. This uses the passed schema for the dataset's schema. That's probably why it isn't catching the incompatibility.

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestFileSystemWriters.java
rdblue: This should fail in Avro because the read schema (with message) and the file schema are incompatible. Avro shouldn't know how to fill in message.

megnataraj:
I am not sure if you got a chance to look into my latest commit addressing all your comments. After adding this new commit, build is failing for CDH5 alone and looking into the logs it doesn't seem those errors were introduced by this new commit, I could be wrong.

chris-hirstein:
@rdblue Have you had a chance to review the update that @megnataraj  provided?

rdblue:
@megnataraj and @chris-hirstein, I've looked at the update and it looks mostly good. My comments on TestAvroWriter#newWriter still applies though. That writer should use the schema that was passed in.

chris-hirstein:
@rdblue It looks like @megnataraj added an update, can you take another look? Thanks!

mkwhitacre:
Re-kicked off Travis build b/c failures seemed unrelated to the PR.

+1 for this PR.

rdblue:
I'm not sure that the tests are exercising the correct code path, so I'd appreciate it if you didn't merge this until I have time to take a closer look. Sorry to be a blocker here.

chris-hirstein:
@rdblue Will you have time to review this change soon?

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestFileSystemWriters.java
```diff
@@ -95,6 +96,43 @@ public void testBasicWrite() throws IOException {
     Assert.assertEquals("Should match written records",
         written, Lists.newArrayList((Iterator) init(reader)));
   }
+  
+  @Test
+  public void testWriteWithOldSchema() throws IOException {
+    Schema writerSchema = SchemaBuilder.record("Message").fields()
```
rdblue: This schema isn't compatible with `TEST_SCHEMA`, which is the dataset's schema. The required string field "message" is missing. This works because you're passing this `writerSchema` into `newReader`.

The problem is that I'm not sure this is testing something well-defined. It validates that this `writerSchema` as passed to `FileSystemWriter.newWriter` is used for the file (otherwise writes would fail) but I don't see the value of using it as the read schema because we expect that it is a valid read schema for both `TEST_SCHEMA` and itself.

I think this should pass the `writerSchema` you have here to `newWriter`, create a file, and validate that the file's schema is actually this one and not `TEST_SCHEMA`. At the writer level, that's what we are interested in. The other behavior (visible/hidden files and other checks) is validated in other tests and can probably be removed.

I also think there needs to be an overall test from the view layer. The motivation is to show that a view with an older write schema works. So I would like to see a view test that creates a dataset with a 2-field schema (one optional field) and writes through a view with a write-compatible schema that is missing the optional field. That validates then end-to-end behavior that we want.

rdblue:
@chris-hirstein, @megnataraj, thanks for the reminder. I've posted a comment about making sure the tests cover what we need to. I hope that is clear enough but if it isn't please ask questions about it and I'll make sure they get answered.

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestFileSystemWriters.java
anandsagark: @rdblue  this unit test is now updated to validate the file's schema is same as writerSchema. 
Also A new test has been added to TestRefinableView, where reader writer compatibility is verified by using two schemas (Value, TestValue). 

Let us know if these are the changes you're expecting, or if there are other modifications that needs to be done

anandsagark:
I've updated this pull request as per the changes suggested by @rdblue 
Please take a look and see if everything looks ok to merge, the failure in the integration checks is not related to this PR

rdblue:
@anandsagark, I'll look today.

rbrush:
+1 overall. I'm willing to merge unless there are objections from @rdblue or @mkwhitacre.

(I might quibble that the TestValue class could do with a more descriptive name, like "ValueWithoutOptionalField", but I don't think that's enough to block it.)

rdblue:
I don't think that the new test actually tests what it should. It looks like the schemas for the two values are the same. I think the dataset should have a 2-field record and the code should write a 1-field record.

But, I've been holding this up for too long and I'm fairly confident that it works, so I'm not going to shoot it down. Thanks for being patient, everyone.

rbrush:
@rdblue Ugh, you're right. Got my wires crossed on which schema the view pointed to. 

@anandsagark, if I'm reading this right I think this just means making testReaderWriterCompatibleSchema use a dataset that uses the "testDescriptor" rather than the "valueDescriptor" so we write to a different schema than the writer has.

anandsagark:
@rdblue  @rbrush  I've update to use different schemas for read, write. Now using 1-field record for write and 2-field record for read. Let me know if this is in tune with what you guys are looking for.

rdblue:
+1

Test looks good to me. Thanks @anandsagark

rbrush:
I kicked off the build again (looks like there was a spurious failure on the default profile). I'll merge pending a clean build on default and CDH5 (I know the CDH4 profile is failing for unrelated reasons.)

rbrush:
The build is clean modulo the known CDH4 issue and we have +1's. Merging.

Thanks everyone!

PR #424, opened 2015-11-30
KITE-1091: Added a moveToTrash method to View, Dataset, Datasets, and…

mkwhitacre:
… DatasetRepository to support safer removal of data. Also converted the Kite DeleteCommand to default to using trash.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
@@ -562,18 +562,7 @@ public class Datasets {
    * @throws IllegalArgumentException if {@code uri} is not a dataset URI
    */
   public static boolean delete(URI uri) {
-    Preconditions.checkArgument(
-        URIBuilder.DATASET_SCHEME.equals(uri.getScheme()),
-        "Not a dataset URI: " + uri);
-
-    Pair<DatasetRepository, Map<String, String>> pair =
-        Registration.lookupDatasetUri(URI.create(uri.getRawSchemeSpecificPart()));
-    DatasetRepository repo = pair.first();
-    Map<String, String> uriOptions = pair.second();
-
-    return repo.delete(
-        uriOptions.get(URIBuilder.NAMESPACE_OPTION),
-        uriOptions.get(URIBuilder.DATASET_NAME_OPTION));
+    return deleteWithTrash(uri, false);
   }
 
   /**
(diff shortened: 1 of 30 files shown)


mkwhitacre:
This will probably have conflict merges with #420 but this has bigger changes so I'll fix up #420 after this gets merged.

mkwhitacre:
@rdblue any advice on how to troubleshoot the Travis build failures? Running the same commands locally and for each profile works so wasn't sure if you had any suggestions on how to more easily troubleshoot or simulate the builds locally.

rdblue:
Sometimes failures that show up in Travis like this are due to running the tests in a different order. Make sure you don't have any dependencies between the tests that would be changed by the order in which they run. Using a TemporaryFolder rule to generate paths is usually a good option to ensure you don't have overlapping datasets.

mkwhitacre:
Yeah I didn't really add new tests in that regard but just added new tests that were the existing delete tests to use the "move to trash".  So might have exposed some ordering problems.

mkwhitacre:
@rdblue got tests fixed by enabling trash for all tests (inconsistent Configuration objects used across tests).  This should be good to review and critique.

rbrush:
+1; looks good to me. Any objections to merging this?

kite-data/kite-data-core/src/test/java/org/kitesdk/data/spi/filesystem/TestFileSystemView.java
```diff
@@ -251,6 +251,147 @@ public void run() {
 
   @Test
   @SuppressWarnings("unchecked")
+  public void testMoveToTrash() throws Exception {
```
rbrush: Is there a practical way to test items are actually moved to the expected trash folder? Perhaps just checking if the presence of directories under .Trash? It looks to me like these tests validate items are removed but not that they actually are in trash.

If this isn't practical in a unit test we can probably just rely on user testing since these at least verify things are removed as expected.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/Datasets.java
```diff
@@ -606,6 +606,75 @@ public static boolean delete(String uri) {
...
+   * URIs must begin with {@code dataset:}. The remainder of
+   * the URI is implementation specific, depending on the dataset scheme.
+   *
+   * @param uri a {@code Dataset} URI
+   * @return {@code true} if any data or metadata is removed, {@code false}
+   *          otherwise
+   * @throws NullPointerException if {@code uri} is null
+   * @throws IllegalArgumentException if {@code uri} is not a dataset URI
+   */
+  public static boolean moveToTrash(URI uri) {
```
rbrush: The logic in this method is a largely a copy of the delete(URI uri) method. Is it worth adding a private delete(URI uri, boolean useTrash) method that both call? I wouldn't really mind leaving it as is, since a small amount of redundant code might be simpler than factoring out a method.

rbrush:
Looks pretty good to me. Just a couple comments/questions above.

mkwhitacre:
@rbrush  thanks for the review  The last commit should address your questions.

If you can give a quick look I'll then squash and rebase to make this merge-able.

mkwhitacre:
Rebased the entire PR to make it merge-able.

rbrush:
Looks like there is an extraneous println in assertDirectoriesInTrash, but +1 otherwise.

mkwhitacre:
Ugh, thanks @rbrush  it has been removed.

rbrush:
+1. (The build failures appear to be unrelated to this change.)

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemUtil.java
```diff
@@ -227,7 +238,7 @@ static boolean cleanlyDelete(FileSystem fs, Path root, Path path) {
           if (stats == null || stats.length == 0) {
             // dir is empty and should be removed
             LOG.debug("Deleting empty path {}", current);
-            deleted = fs.delete(current, true) || deleted;
+            deleted = (useTrash ? Trash.moveToAppropriateTrash(fs, current, fs.getConf()) : fs.delete(current, true)) || deleted;
```
noslowerdna: I don't think moving to trash is appropriate here. We know that a child directory of the pending-trashed directory is already in trash (assuming deleted = true as of line 231), so moving a now-empty parent directory would just create a timestamp-suffixed duplicate empty directory in trash, at each level of the path hierarchy. Reverting back to the original code here should be completely safe and a lot cleaner.

mkwhitacre:
@noslowerdna I removed the recursive trash and went with the delete.  I did also rebase to a single commit so it is a cleaner merge.

noslowerdna:
+1

mkwhitacre:
Thanks @noslowerdna merging.

PR #431, opened 2016-01-23
KITE-950 Parse entity with external column mapping.

nethy:


jarcec:
It seems that travis build has failed, could you take a look if it's relevant Gabor?

nethy:
It says the build was stalling for 10 minutes, that's why it's failed. Tomorrow I'll try to execute all the tests again and give a more precise feedback.

nethy:
I've just run the build locally for the whole kite project and it was successful, but it took more than 16 minutes to finish.
Can it be the CI hasn't enough resource in that moment and it caused the build stalling?
I would run the build again, but I do not know there is a chance to do that without a new pull request.

PR #432, opened 2016-01-27
KITE-1102: Correct URI pattern registrations for s3a protocol

noslowerdna:
Minor change to fix https://issues.cloudera.org/browse/KITE-1102


rdblue:
+1

I believe we left it as it was because we recommend S3N. Originally, we weren't sure whether the *path component would be useful to S3 users because buckets are available as an alternative.

mkwhitacre:
+1

mkwhitacre:
Can ignore the cdh4 profile failure b/c that is related to flume-ng download.  Rekicked off the default profile however as that seemed like a hung test.

mkwhitacre:
Build successful aside from flaky cdh4 build.  merging.

PR #434, opened 2016-02-09
KITE-1101: Added logic to FileSystemPartitionIterator to exclude dire…

mkwhitacre:
…ctories with data that does not qualify. Extended predicates and StorageKey code to support partial values.

Note this might violate some of the initial designs for StorageKey and the Predicates so I'm open to alternate ways of doing this logic.  @rdblue and @rbrush if you have some insight or critique that would be helpful.  It is not necessarily the cleanest changes so if you have suggestions on how to clean this up with original design intent that would be helpful.

In some smaller tests I've seen significant improvements in job startup time (1s vs 100s) and a lot less memory pressure on the driver program when kicking off the job.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/StorageKey.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/StorageKey.java
@@ -96,12 +96,15 @@ public class StorageKey extends Marker implements Comparable<StorageKey> {
 
   @Override
   public boolean has(String name) {
-    return fields.containsKey(name);
+    //check if id value exists and it is less than the size meaning we have a value.
+    Integer id = fields.get(name);
+    return id != null && id < values.size();
   }
 
   @Override
   public Object get(String name) {
-    return values.get(fields.get(name));
+    int index = fields.get(name);
+    return index < values.size() ? values.get(index) : null;
   }
 
   /**
@@ -132,8 +135,6 @@ public class StorageKey extends Marker implements Comparable<StorageKey> {
    */
   public void replaceValues(List<Object> values) {
     Preconditions.checkArgument(values != null, "Values cannot be null");
-    Preconditions.checkArgument(values.size() == fields.size(),
-        "Not enough values for a complete StorageKey");
     this.values = values;
   }
 
@@ -182,6 +183,7 @@ public class StorageKey extends Marker implements Comparable<StorageKey> {
   public StorageKey reuseFor(Path path, PathConversion conversion) {
     conversion.toKey(path, this);
     this.path = path;
+
     return this;
   }
 
@@ -204,6 +206,10 @@ public class StorageKey extends Marker implements Comparable<StorageKey> {
       }
     }
     return 0;
+  }
+
+  public int size(){
+    return values.size();
   }
 
   @Override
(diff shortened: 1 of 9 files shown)


kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/StorageKey.java
```diff
@@ -132,8 +135,6 @@ public void replace(int index, Object value) {
    */
   public void replaceValues(List<Object> values) {
     Preconditions.checkArgument(values != null, "Values cannot be null");
-    Preconditions.checkArgument(values.size() == fields.size(),
```
noslowerdna: Would it make sense to change this to <= as a sanity check?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemPartitionIterator.java
```diff
@@ -62,15 +66,28 @@ public FileSystemIterator(int depth) throws IOException {
...
+      if(key.size() > 0){
+        applies = predicate.apply(key);
+      }else{
+        LOG.debug("Not applying predicate proactively because path {} does not have any key values.", dir);
+      }
+      if(applies) {
+        try {
+          for (FileStatus stat : fs.listStatus(dir, PathFilters.notHidden())) {
+            if (stat.isDir()) {
+              // TODO: add a check here for range.couldContain(Marker)
```
noslowerdna: Is this TODO comment still applicable?

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemViewKeyInputFormat.java
```diff
@@ -19,6 +19,7 @@
 import com.google.common.collect.Lists;
 import java.io.IOException;
 import java.util.Iterator;
+import java.util.LinkedList;
```
noslowerdna: Looks like you can revert the changes in this file assuming all unintentional.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemPartitionIterator.java
mkwhitacre: That's a good question.  That TODO was there prior to the changes and I'm not sure I completely grokked the optimization.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemViewKeyInputFormat.java
mkwhitacre: Yep, I put some timings in this class and didn't get them cleanly removed.  I'll revert this class changes.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/StorageKey.java
mkwhitacre: Follow up commit should have this sanity check.

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemViewKeyInputFormat.java
mkwhitacre: Rolled back this file.

noslowerdna:
+1

rbrush:
+1. Looks good to me.

mkwhitacre:
Got two +1s.  Merging.

Issue #435, opened 2016-02-17
extractAvroPaths does not traverses arrays with the '[]' notation when array is part of a union


SOLID-Design:
When avro schema contains an array within a union then the '[]' operator does not traverses the array. 

Given schema:
`{ "namespace": "org.kitesdk.morphline.avro",
    "type": "record",
    "name": "ArrayInUnionTestRecord",
    "fields": [
          {"name": "items", "type": {"type": "array", "items": "string"}},
          {"name": "itemsInUnion", "type": ["null", {"type": "array", "items":"string"}], "default": null}
     ]
  }`

and configuration
`extractAvroPaths {
       paths : {
            "/items[]" : "/items[]"
            "/itemsInUnion[]" : "/itemsInUnion[]"
       }
 }`

_/items[]_ is properly extracted but _/itemsInUnion[]_ is not.

whoschek:
Thanks for the contribution! I committed a fix that pinpoints the issue a bit more precisely, here: https://github.com/kite-sdk/kite/commit/50781c4ddca9763304119185d6caf86f6b1f6ef1

PR #437, opened 2016-02-18
fix for extractAvroPaths traversal of arrays within union

SOLID-Design:
fixes issue #435 - extractAvroPaths does not traverses arrays
with the '[]' notation when array is part of a union


whoschek:
Thanks for the contribution! I committed a fix that pinpoints the issue a bit more precisely, here: https://github.com/kite-sdk/kite/commit/50781c4ddca9763304119185d6caf86f6b1f6ef1

PR #438, opened 2016-02-21
KITE-1114: Kite CLI json-import HDFS temp file path not multiuser safe

akalaszi:


jarcec:
I don't consider the failure in CDH4 relevant nor important - CDH4 has been EOLed long time ago.

PR #440, opened 2016-03-03
expose the current Command name to the CommandBuilder instance

leonardofoderaro:
Hi,
thank you for this great library.
I'd need the current Command name in the public Command build method.
Since a Command can be invoked by multiple names it could be useful to save the current name used to invoke it in order to make it available to the CommandBuilder, possibly without modify the Command Interface (as it would impact all the existing commands).
Is there a better way to get this information instead of my suggested patch?
Thank you,
Leonardo


whoschek:
The AbstractCommand ctor already has a CommandBuilder param and CommandBuilder.names() would give you the names used for this command (although not the actually used name in case there are multiple aliases). Would that work for you?

PR #442, opened 2016-03-04
KITE-1118: Optimize writing data to S3

noslowerdna:
Pull request for https://issues.cloudera.org/browse/KITE-1118

For importing about 20 GB of data into S3, this provided approximately an 8x job speedup, from 40 min to 5 min.

### DIFF
--- a/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemProperties.java
+++ b/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemProperties.java
@@ -60,4 +60,12 @@ public class FileSystemProperties {
    * be closed and finalized once they reach this age.
    */
   public static final String ROLL_INTERVAL_S_PROP = "kite.writer.roll-interval-seconds";
+
+  /**
+   * Until HADOOP-9565 is available and fully adopted, need to make this configurable so that
+   * we can avoid multiple expensive copy operations when writing output to a file system that
+   * does not support fast file renaming. Default is false unless the FileSystem URI scheme is
+   * s3n or s3a.
+   */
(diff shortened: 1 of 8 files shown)


mkwhitacre:
+1

kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/filesystem/FileSystemProperties.java
```diff
@@ -60,4 +60,12 @@
...
    */
   public static final String ROLL_INTERVAL_S_PROP = "kite.writer.roll-interval-seconds";
+
+  /**
+   * Until HADOOP-9565 is available and fully adopted, need to make this configurable so that
+   * we can avoid multiple expensive copy operations when writing output to a file system that
+   * does not support fast file renaming. Default is false unless the FileSystem URI scheme is
+   * s3n or s3a.
+   */
+  public static final String OBJECTSTORE_FILESYSTEM = "kite.writer.object-store";
```
rbrush: Just one suggestion here: it seems like this decision should be based on whether the filesystem supports efficient rename operations, since future object stores may support rename (and s3 may get it some day). Perhaps this setting should be "kite.writer.supports-rename", and by default we assume that s3a and s3n filesystems do not?

I think this keeps Kite focused on the properties it's interested in...where it shouldn't really care or make assumptions about the properties of an object store.

rbrush:
Made one suggestion above, but otherwise this looks good.

noslowerdna:
@rbrush I like that suggestion, changed: https://github.com/noslowerdna/kite/commit/8b28ae7f4187871a94c919d2f8b68514b4287a2b

Thanks for the review.

rbrush:
+1, looks great!

mkwhitacre:
Re-kicked off the "default" profile build.

mkwhitacre:
Build passed.  Merging.

PR #443, opened 2016-04-15
Morphlines hashDigest command

tmgstevens:
Initial commit for hashCommand implementation and tests

### DIFF
--- a/kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashDigestBuilder.java
+++ b/kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashDigestBuilder.java
@@ -1,0 +1,220 @@
+
+/*
+ * Copyright 2013 Cloudera Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kitesdk.morphline.stdlib;
+
+import java.nio.charset.Charset;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.ListIterator;
+
+import org.kitesdk.morphline.api.Command;
+import org.kitesdk.morphline.api.CommandBuilder;
+import org.kitesdk.morphline.api.MorphlineCompilationException;
+import org.kitesdk.morphline.api.MorphlineContext;
+import org.kitesdk.morphline.api.Record;
+import org.kitesdk.morphline.base.AbstractCommand;
+
+import com.google.common.base.Charsets;
+import com.typesafe.config.Config;
+
+/**
+ * Command that creates a hash digest of a field so-as to deterministically
+ * provide a unique key
+ */
+
+public class HashDigestBuilder implements CommandBuilder {
+
+  @Override
+  public Collection<String> getNames() {
+    return Collections.singletonList("hashDigest");
+  }
+
+  @Override
+  public Command build(Config config, Command parent, Command child,
+      MorphlineContext context) {
+    return new HashDigest(this, config, parent, child, context);
+  }
+
+  ///////////////////////////////////////////////////////////////////////////////
+  // Nested classes:
+  ///////////////////////////////////////////////////////////////////////////////
+  private static final class HashDigest extends AbstractCommand {
+
+    private final String inputFieldName;
+    private final String outputFieldName;
+    private final String hashType;
+    private final boolean preserveExisting;
+    private final Charset charset;
+    private final MessageDigest digest;
+
+    private static final String INPUT_FIELD = "inputField";
+    private static final String OUTPUT_FIELD = "outputField";
+    private static final String HASH_TYPE = "hashType";
+    private static final String PRESERVE_EXISTING_NAME = "preserveExisting";
+    private static final String CHARSET_FIELD = "charset";
+    private static final boolean PRESERVE_EXISTING_DEFAULT = true;
+
+    public HashDigest(CommandBuilder builder, Config config, Command parent,
+        Command child, MorphlineContext context) {
+      super(builder, config, parent, child, context);
+
+      this.inputFieldName = getConfigs().getString(config, INPUT_FIELD);
+      this.outputFieldName = getConfigs().getString(config, OUTPUT_FIELD);
+      this.hashType = getConfigs().getString(config, HASH_TYPE);
+      this.preserveExisting = getConfigs().getBoolean(config,
+          PRESERVE_EXISTING_NAME, PRESERVE_EXISTING_DEFAULT);
+      this.charset = getConfigs().getCharset(config, CHARSET_FIELD, Charsets.UTF_8);
+      
+
+      try {
+        this.digest = MessageDigest.getInstance(hashType);
+      } catch (NoSuchAlgorithmException e) {
+        throw new MorphlineCompilationException("Unable to initialise digest", config, e);
+      }
+      
+      validateArguments();
+
+      if (LOG.isTraceEnabled()) {
+        LOG.trace("inputField: {}", inputFieldName);
+        LOG.trace("outputField: {}", outputFieldName);
+        LOG.trace("hashType: {}", hashType);
+        LOG.trace("preserveExisting: {}", preserveExisting );
+      }
+    }
+
+    @Override
+    protected boolean doProcess(Record record) {
+
+      if (!preserveExisting) {
+        record.removeAll(outputFieldName);
+      }
+
+      ListIterator iter = record.get(inputFieldName).listIterator();
+      
+      while (iter.hasNext()) {
+        Object inputField = iter.next();
+        byte[] inputFieldBytes;
+        if (inputField == null ) {
+          inputFieldBytes = null;
+        } else if (inputField instanceof byte[]) {
+          inputFieldBytes = (byte[])inputField;
+        } else {
+          inputFieldBytes = inputField.toString().getBytes(charset);
+        }
+        record.put(outputFieldName, doHash(inputFieldBytes));
+      }
+
+      // pass record to next command in chain:
+      return super.doProcess(record);
+
+    }
+
+    private String doHash(byte[] inputBytes) {
+      digest.reset();
+      
+      return Hex.encodeHexString(digest.digest(inputBytes));
+    }
+
+  }
+
+  //Manually shaded subset from org.apache.commons.codec 1.8.
+  /*
+   * Licensed to the Apache Software Foundation (ASF) under one or more
+   * contributor license agreements.  See the NOTICE file distributed with
+   * this work for additional information regarding copyright ownership.
+   * The ASF licenses this file to You under the Apache License, Version 2.0
+   * (the "License"); you may not use this file except in compliance with
+   * the License.  You may obtain a copy of the License at
+   *
+   *      http://www.apache.org/licenses/LICENSE-2.0
+   *
+   * Unless required by applicable law or agreed to in writing, software
+   * distributed under the License is distributed on an "AS IS" BASIS,
+   * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   * See the License for the specific language governing permissions and
+   * limitations under the License.
+   */
+  private static class Hex {
+    /**
+     * Used to build output as Hex
+     */
+    private static final char[] DIGITS_LOWER =
+        {'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'};
+
+    /**
+     * Used to build output as Hex
+     */
+    private static final char[] DIGITS_UPPER =
+        {'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F'};
+
+    
+    protected static char[] encodeHex(final byte[] data, final char[] toDigits) {
+      final int l = data.length;
+      final char[] out = new char[l << 1];
+      // two characters form the hex value.
+      for (int i = 0, j = 0; i < l; i++) {
+          out[j++] = toDigits[(0xF0 & data[i]) >>> 4];
+          out[j++] = toDigits[0x0F & data[i]];
+      }
+      return out;
+    }
+    /**
+     * Converts an array of bytes into a String representing the hexadecimal values of each byte in order. The returned
+     * String will be double the length of the passed array, as it takes two characters to represent any given byte.
+     *
+     * @param data
+     *            a byte[] to convert to Hex characters
+     * @return A String containing hexadecimal characters
+     * @since 1.4
+     */
+    public static String encodeHexString(final byte[] data) {
+        return new String(encodeHex(data));
+    }
+    
+    /**
+     * Converts an array of bytes into an array of characters representing the hexadecimal values of each byte in order.
+     * The returned array will be double the length of the passed array, as it takes two characters to represent any
+     * given byte.
+     *
+     * @param data
+     *            a byte[] to convert to Hex characters
+     * @return A char[] containing hexadecimal characters
+     */
+    public static char[] encodeHex(final byte[] data) {
+        return encodeHex(data, true);
+    }
+    
+    /**
+     * Converts an array of bytes into an array of characters representing the hexadecimal values of each byte in order.
+     * The returned array will be double the length of the passed array, as it takes two characters to represent any
+     * given byte.
+     *
+     * @param data
+     *            a byte[] to convert to Hex characters
+     * @param toLowerCase
+     *            <code>true</code> converts to lowercase, <code>false</code> to uppercase
+     * @return A char[] containing hexadecimal characters
+     * @since 1.4
+     */
+    public static char[] encodeHex(final byte[] data, final boolean toLowerCase) {
+        return encodeHex(data, toLowerCase ? DIGITS_LOWER : DIGITS_UPPER);
+    }
+    
+  }
+}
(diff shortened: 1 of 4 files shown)


tmgstevens:
Link to https://issues.cloudera.org/browse/KITE-1122

kite-morphlines/kite-morphlines-core/pom.xml
```diff
@@ -62,6 +62,11 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+    	<groupId>commons-codec</groupId>
+    	<artifactId>commons-codec</artifactId>
```
whoschek: The policy for existing morphline mvn modules is to not add additional dependencies on further 3rd party libs, in order to avoid the potential for dependency version conflicts / jar hell. Consider providing similar functionality just using what's in the JDK or with existing dependencies, or consider adding a new separate mvn module, e.g. along the lines of the kite-morphlines-json module.

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashCommandBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+ * Command that converts the timestamps in a given field from one of a set of
+ * input date formats (in an input timezone) to an output date format (in an
+ * output timezone), while respecting daylight savings time rules. Provides
+ * reasonable defaults for common use cases.
+ */
+
+public class HashCommandBuilder implements CommandBuilder {
+
+	@Override
+	public Collection<String> getNames() {
```
whoschek: indents in kite are via 2 spaces by convention.

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashCommandBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+			MessageDigest digest = DigestUtils.getDigest(hashType);
+
+			byte[] inputBytes = StringUtils.getBytesUtf8(inputString);
+
+			return Hex.encodeHexString(digest.digest(inputBytes));
+		}
+
+		@Override
+		protected void doNotify(Record notification) {
+			LOG.debug("myNotification: {}", notification);
```
whoschek: is this debug msg necessary?

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashCommandBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.commons.codec.digest.DigestUtils;
+import org.kitesdk.morphline.api.Command;
+import org.kitesdk.morphline.api.CommandBuilder;
+import org.kitesdk.morphline.api.MorphlineContext;
+import org.kitesdk.morphline.api.Record;
+import org.kitesdk.morphline.base.AbstractCommand;
+import com.typesafe.config.Config;
+
+/**
+ * Command that converts the timestamps in a given field from one of a set of
```
whoschek: pls update doc from copy n pasted timestamp based doc

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashCommandBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+	// Nested classes:
+	///////////////////////////////////////////////////////////////////////////////
+	private static final class HashDigest extends AbstractCommand {
+
+		private final String inputFieldName;
+		private final String outputFieldName;
+		private final String hashType;
+		private final boolean preserveExisting;
+
+		public static final String INPUT_FIELD = "inputField";
```
whoschek: making these names public won't help much because the HashDigest class itself is private. The names are effectively part of the public API and "set in stone" because changing the names after the release would break existing client morphline configs.

kite-morphlines/kite-morphlines-core/src/test/resources/test-morphlines/hashCommand.conf
```diff
@@ -0,0 +1,52 @@
...
+# limitations under the License.
+
+morphlines : [
+  {
+    id : morphline1
+    importCommands : ["org.kitesdk.**", "org.cloudera.tristan.morphlines.**"]
+    
+    commands : [    
+    { logDebug { format : "input record: {}", args : ["@{}"] } }                 
+      {
```
whoschek: remove indent for consistency

kite-morphlines/kite-morphlines-core/src/test/resources/test-morphlines/hashCommand.conf
```diff
@@ -0,0 +1,52 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+morphlines : [
+  {
+    id : morphline1
+    importCommands : ["org.kitesdk.**", "org.cloudera.tristan.morphlines.**"]
```
whoschek: org.cloudera.tristan.morphlines seems outdated wrt. the package name of the submitted HashDigestBuilder class

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashCommandBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+import com.typesafe.config.Config;
+
+/**
+ * Command that converts the timestamps in a given field from one of a set of
+ * input date formats (in an input timezone) to an output date format (in an
+ * output timezone), while respecting daylight savings time rules. Provides
+ * reasonable defaults for common use cases.
+ */
+
+public class HashCommandBuilder implements CommandBuilder {
```
whoschek: Builder class should be named "HashDigestBuilder" if the command is named "hashDigest" per convention. Or maybe "GenerateHashDigestBuilder" with command named "generateHashDigest" ala "generateUUID" command?

whoschek:
What's the probability of hash collision (two different messages returning the same id) with the MessageDigests vs. GenerateUUID impl? How much less likely is a collision with SHA-256 than with that we have already today?

tmgstevens:
Regarding comparison vs GenerateUUID - the latter is guaranteed to be unique regardless of the input document - i.e. running the same document through a billion times will give a billion UUIDs.

This new function will always return the same result for the same input - hence its desirability.

Whether a collision is likely depends on the algorithm chosen and the number of inputs, but casually speaking is close to zero :-)

kite-morphlines/kite-morphlines-core/pom.xml
tmgstevens: Resolved by shading the relevant methods into a static class. Shading the whole classes had too many dependencies.

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashCommandBuilder.java
tmgstevens: Resolved

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashCommandBuilder.java
tmgstevens: Removed

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashCommandBuilder.java
tmgstevens: Updated

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashCommandBuilder.java
tmgstevens: GPWM. Updated.

kite-morphlines/kite-morphlines-core/src/test/resources/test-morphlines/hashCommand.conf
tmgstevens: That's embarrassing! Resolved.

kite-morphlines/kite-morphlines-core/src/test/resources/test-morphlines/hashCommand.conf
tmgstevens: That's embarassing! Resovled

tmgstevens:
Thanks for reviewing @whoschek - I've updated based on your feedback.

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashDigestBuilder.java
```diff
@@ -0,0 +1,215 @@
...
+    private String doHash(String inputString) {
+      MessageDigest digest = DigestUtils.getDigest(hashType);
+
+      byte[] inputBytes = StringUtils.getBytesUtf8(inputString);
+
+      return Hex.encodeHexString(digest.digest(inputBytes));
+    }
+
+    @Override
+    protected void doNotify(Record notification) {
```
whoschek: now doNotify() nomore does anything and the method can be removed, which improves readability.

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashDigestBuilder.java
```diff
@@ -0,0 +1,215 @@
...
+    protected boolean doProcess(Record record) {
+
+      if (!preserveExisting) {
+        record.removeAll(outputFieldName);
+      }
+
+      ListIterator iter = record.get(inputFieldName).listIterator();
+      
+      while (iter.hasNext()) {
+        record.put(outputFieldName, doHash(iter.next().toString()));
```
whoschek: How about adding a bit of logic that makes it possible for the user to also pass a byte[], without conversion from string to byte[]? This enables the user to use any algorithm to convert input data to a byte[] (not just UTF8 string conversion), in a prior morphline command. Along these lines:

Object next = iter.next();
byte[] bytes;
if (next instanceof byte[]) { 
  bytes =  (byte[]) next;
} else { 
  bytes = next.toString();
}
... hex encode (bytes)

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashDigestBuilder.java
```diff
@@ -0,0 +1,215 @@
...
+    @Override
+    protected void doNotify(Record notification) {
+      super.doNotify(notification);
+    }
+
+  }
+  
+  
+  //Manually shaded from org.apache.commons.codec 1.10.
+  public static class DigestUtils {
```
whoschek: I think you can just delete the entire DigestUtils class and simply paste the getDigest(hashType) one liner into your doHash() method. This removes unnecessary dead code and improves maintainability.

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashDigestBuilder.java
```diff
@@ -0,0 +1,215 @@
...
+        try {
+            return MessageDigest.getInstance(algorithm);
+        } catch (final NoSuchAlgorithmException e) {
+            throw new IllegalArgumentException(e);
+        }
+    }
+  }
+
+  //Manually shaded from org.apache.commons.codec 1.8.
+  public static class Hex {
```
whoschek: let's declare that Hex class as "private static class" to improve clarity.

kite-morphlines/kite-morphlines-core/src/test/resources/test-morphlines/hashCommand.conf
```diff
@@ -15,10 +15,11 @@
 morphlines : [
   {
     id : morphline1
-    importCommands : ["org.kitesdk.**", "org.cloudera.tristan.morphlines.**"]
+    importCommands : ["org.kitesdk.**"]
     
     commands : [    
-    { logDebug { format : "input record: {}", args : ["@{}"] } }                 
+    { 
+      logDebug { format : "input record: {}", args : ["@{}"] } }
```
whoschek: Can we move the closing brace } to the same indentation level as the opening brace { ?

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashDigestBuilder.java
whoschek: Also please copy the original Apache licene blurb from the original Hex.java file into a location just before the shaded Hex code. This clarifies things.

whoschek: Also for better perf, might be good to move the the MessageDigest.getInstance() call into the ctor of the morphline command and on each invocation of doProcess() we'd reuse the instance via MessageDigest.reset(). This also has the nice side effect of failing fast in the command ctor if the algorithm id is illegal or unknown.

whoschek:
Also, let's rename the name of the test case class to HashDigestTest in order to reflect the "hashDigest" name.

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashDigestBuilder.java
tmgstevens: Removed

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashDigestBuilder.java
tmgstevens: +1 Done

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashDigestBuilder.java
tmgstevens: Done, apart from I've used StringUtils.getBytesUtf8 to avoid doing the charset lookup for each string.

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashDigestBuilder.java
tmgstevens: I'd thought that MessageDigest wasn't re-usable, but having researched that is not the case.

kite-morphlines/kite-morphlines-core/src/test/resources/test-morphlines/hashCommand.conf
tmgstevens: Done

kite-morphlines/kite-morphlines-core/src/test/resources/test-morphlines/hashCommand.conf
```diff
@@ -0,0 +1,73 @@
...
+    commands : [    
+      
+      { logDebug { format : "input record: {}", args : ["@{}"] } }                 
+      
+      { 
+        readBlob {}
+      } 
+      
+      {
+        hashDigest {
```
whoschek: Can we rename hashCommand.conf to hashDigest.conf to be consistent with the previous renames?

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashDigestBuilder.java
```diff
@@ -0,0 +1,213 @@
...
+
+      ListIterator iter = record.get(inputFieldName).listIterator();
+      
+      while (iter.hasNext()) {
+        Object inputField = iter.next();
+        byte[] inputFieldBytes;
+        if (inputField instanceof byte[]) {
+          inputFieldBytes = (byte[])inputField;
+        } else {
+          inputFieldBytes = StringUtils.getBytesUtf8(inputField.toString());
```
whoschek: Can we just manually inline that one-liner: inputField.toString().getBytes(Charsets.UTF_8) and revert the StringUtils class back to non-public? That way StringUtils can't possibly be mistaken for a public API.

whoschek:
Thanks again for the great contrib! I think we're almost there. After the final minor outstanding changes please squash the multiple commits into a single big commit, and then we're ready to merge.

whoschek:
Also would be good to add corresponding docs in kite-morphlines/src/site/confluence/morphlinesReferenceGuide.confluence but that can be done in a separate commit if you like.

whoschek:
To read the html output that is generated from the confluence docs, or for debugging: cd kite; mvn post-site -pl kite-morphlines; open kite-morphlines/target/site/index.html

whoschek:
Also, for better perf consider using UTF16 instead of UTF8 for string conversion.

kite-morphlines/kite-morphlines-core/src/test/resources/test-morphlines/hashCommand.conf
tmgstevens: Fixed

kite-morphlines/kite-morphlines-core/src/main/java/org/kitesdk/morphline/stdlib/HashDigestBuilder.java
tmgstevens: Yes - I had originally thought there was more complexity in that method. I've also added a check for null to prevent NPEs.

tmgstevens:
Regarding using UTF16 rather than UTF8 - this would give different results as the input byte array would be different. I think in general people would expect UTF8 (for example http://www.miraclesalad.com/webtools/md5.php) however actually we can make that configurable fairly simply. I'll set the default to UTF-8 but add a parameter call charset as per readCSV command.

tmgstevens:
Thanks for reviewing @whoschek  - much appreciated.

tmgstevens:
Updated the docs onto the same commit. Let me know if there's any further changes required.

whoschek:
Hi Tristan. I merged this. Thanks for this great contrib!

tmgstevens:
Thanks Wolfgang for your feedback and swift turnaround.
Tristan

Tristan Stevens
Senior Solutions Architect
Cloudera, Inc. | www.cloudera.com

m +44(0)7808 986422 |

PR #466, opened 2017-02-09
KITE-1137 update javaVersion to 1.7 and fix warnings

szantaikis:
* according to the readme default should be java 7 so might as well update the target version property
 * In top level pom.xml the maven compiler plugin to use the javaVersion property instead of the hard coded 1.6 value
 * Some warnings had to be fixed in HiveUtils, because of the -Werror (error on warnings) maven flag

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,8 +76,8 @@ For Hadoop 1:
 mvn install -Dhadoop.profile=1
 ```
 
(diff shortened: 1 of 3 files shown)


whoschek:
Thanks for the great contrib, Peter!

PR #467, opened 2017-02-09
KITE-1136 SolrLocator updated to be able to use managed schemas + unit tests

szantaikis:


whoschek:
Thanks for this great contrib, Peter!

PR #476, opened 2017-08-18
HiveAbstractDatasetRepository now uses the first URI from hive.metast…

szvasas:
…ore.uris property.


QwertyManiac:
Thank you for unblocking this scenario!

Why not pass both the addresses onwards to the HMS client built later, to allow it to use HMS HA normally?

Sorry for coming in late - only just saw the JIRA update.

szvasas:
Hi Harsh,

Implementing the HMS HA support in Kite and testing it properly with Sqoop would have taken a bigger effort and we needed to resolve this bug ASAP. 
The HMS HA support is on our roadmap as well and it is going to be implemented in another patch.

Szabolcs

PR #485, opened 2018-11-14
Add Google Cloud Storage Dataset Support

devlucasc:


devlucasc:
I think that has a problem in travis with JDK7 that makes the build crashes.

"The JAVA_HOME environment variable is not defined correctly
This environment variable is needed to run this program
NB: JAVA_HOME should point to a JDK not a JRE"

